PR #84, opened 2023-09-20
Some more fixes

s-hemer:
First three commits are considered ready to merge: Fixes html path in README, the libyaml install hint and restructures the standalone kernel build instructions (mainly to add the hint to build the DTs). 
The remaining three are RFC/draft for nicer console blocks. One additional idea, that I did not investigate yet: we may introduce a `.. code-block:: host_console` that automatically adds the `host:$` (maybe colored, we discussed this at FAE that people do not get the host/target hint, so it may help to have different colors for both?!)


jonas-rem:
@s-hemer I will close this PR for now to maintain a better overview of the repository and active PRs. Feel free to re-open anytime!

PR #145, opened 2024-01-29
Wip/bhahn/bspimx8 m 3123

BHahn42618:
split development chapter
try to resolve confusion about which development manifest is meant. (RFC)


ymoog (dismissed):

source/bsp/imx8/development/development_manifests.rsti
```diff
@@ -1,8 +1,8 @@
 Accessing the Development States between Releases
```
tremmet, line 1: As discussed please move "Development States between Releases" in a sub chapter and also create a new sub chapter for the upcoming release.

tremmet (changes requested):
Only the change for the manifest checkout of dev states. Rest looks good.

jremmert-phytec-iot:
The file `development.rsti` is not required anymore in this case right? Could you delete it in this PR in this case?

source/bsp/imx8/development/development_manifests.rsti
```diff
@@ -1,8 +1,11 @@
...
--------------------------------------------------
+Accessing the Development states
+--------------------------------
 
-Special release manifests exist to give you access to the current development
-states of the *Yocto* BSP. They will not be displayed in the phyLinux selection
+Development state of current release
+....................................
+
+Special release manifests exist to give you access to the development states
```
tremmet, line 7: Please remove the term `Special` here. As we also have `Special Releases` so this might be confusing.

tremmet (changes requested):

source/bsp/imx8/development/development_manifests.rsti
BHahn42618: Done

tremmet (approved):

ymoog (approved):

PR #146, opened 2024-01-31
BSPIMX8M-3139: create PD24.1.0.rst BSP manual

ymoog:
For upcoming imx8mp mainline BSP release, add a manual.
This is done because the content and structure may be different from vendor releases. Since the plan is to move back to vendor release after the mainline release, we do not work on the head manual and branch off early.

Add some misc. improvements that I noticed during proofreading.

Depends on #145

### DIFF
--- a/source/bsp/getting-started.rsti
+++ b/source/bsp/getting-started.rsti
@@ -11,7 +11,7 @@ There are several ways to flash an image to SD card or even eMMC. Most notably
 using simple, sequential writing with the Linux command line tool ``dd``. An
 alternative way is to use PHYTEC's system initialization program called
 `partup <https://github.com/phytec/partup>`_, which makes it especially easy to
-format more complex systems. You can get `prebuild Linux binaries of partup
+format more complex systems. You can get `prebuilt Linux binaries of partup
 <https://github.com/phytec/partup/releases>`__ from its release page. Also read
 `partup's README <https://github.com/phytec/partup#readme>`__ for installation
 instructions.
@@ -93,8 +93,8 @@ Unmount all partitions, e.g.:
 
    host:~$ sudo umount /dev/sde1
 
-Now, the SD card is ready to be flashed with an image, using either partup,
-``dd`` or bmap-tools.
+Now, the SD card is ready to be flashed with an image, using either ``partup``,
+``dd`` or ``bmap-tools``.
 
 Using partup
 ............
@@ -124,8 +124,8 @@ Further usage of partup is explained at its `official documentation website
 Using bmap-tools
 ................
 
-An alternative and also fast way to prepare an SD card can be done by using
-`bmap-tools <https://github.com/intel/bmap-tools>`_ from Intel. Yocto
+An alternative and also fast way to prepare an SD card is using
+`bmap-tools <https://github.com/intel/bmap-tools>`_ by Intel. Yocto
 automatically creates a block map file (``<IMAGENAME>-<MACHINE>.wic.bmap``) for
 the WIC image that describes the image content and includes checksums for data
 integrity. *bmaptool* is packaged by various Linux distributions. For
--- a/source/bsp/imx8/imx8mp/imx8mp.rst
+++ b/source/bsp/imx8/imx8mp/imx8mp.rst
@@ -12,6 +12,16 @@ HEAD
 
    head
 
+Mainline HEAD
+=============
+
+.. toctree::
+   :caption: Table of Contents
+   :numbered:
+   :maxdepth: 2
+
+   mainline-head
+
 PD23.1.0
 ========
 
--- a/source/bsp/imx8/imx8mp/mainline-head.rst
+++ b/source/bsp/imx8/imx8mp/mainline-head.rst
@@ -1,0 +1,446 @@
+.. Download links
+.. |dlpage-bsp| replace:: our BSP
+.. _dlpage-bsp: https://www.phytec.de/bsp-download/?bsp=BSP-Yocto-NXP-i.MX8MP-PD23.1.0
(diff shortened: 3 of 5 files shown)


source/bsp/getting-started.rsti
```diff
@@ -11,7 +11,7 @@ There are several ways to flash an image to SD card or even eMMC. Most notably
 using simple, sequential writing with the Linux command line tool ``dd``. An
 alternative way is to use PHYTEC's system initialization program called
 `partup <https://github.com/phytec/partup>`_, which makes it especially easy to
-format more complex systems. You can get `prebuild Linux binaries of partup
+format more complex systems. You can get `prebuilt Linux binaries of partup
```
tremmet, line 14: would be "pre-built" even more correct?

tremmet (approved):

source/bsp/getting-started.rsti
ymoog: I don't know tbh. [Apparently not](https://dictionary.cambridge.org/dictionary/english/prebuilt)

jremmert-phytec-iot (changes requested):
I think that customers could be confused by thinking that PD24.1.0 has been released already. Can we make this more obvious that is an unreleased BSP? Remember that everything gets published immediately, which was not the case before with our confluence setting.

Like e.g.:

- Adding a red warning box at the beginning of the manual. Maybe we can even refer to a doc that explains how customers could checkout this WIP BSP for testing? After all, whenever the reader consult Docs for released BSPs, they see those new and unreleased BSPs in the sidebar and that might raise their interest (which is a good thing I guess).
- Changing the manual Name to PD24.1.y from PD24.1.0. Once we release the manual we can change this.

What do you think?

ymoog:
> I think that customers could be confused by thinking that PD24.1.0 has been released already. Can we make this more obvious that is an unreleased BSP? Remember that everything gets published immediately, which was not the case before with our confluence setting.
> 
> Like e.g.:
> 
>     * Adding a red warning box at the beginning of the manual. Maybe we can even refer to a doc that explains how customers could checkout this WIP BSP for testing? After all, whenever the reader consult Docs for released BSPs, they see those new and unreleased BSPs in the sidebar and that might raise their interest (which is a good thing I guess).
> 
>     * Changing the manual Name to PD24.1.y from PD24.1.0. Once we release the manual we can change this.
> 
> 
> What do you think?

I agree it is a challenge to avoid confusion.

My suggestion is the following:
1. Create "feature" branch where all PR are merged to.
2. Use normal workflow for working on pd24.1.0
3. Merge feature branch into main -> tag it.

Then, no pd24 will be public at all. We have a pd24 draft that we can continously develop until release.
After release the feature branch is deleted so no complexity regarding multiple branches.

jremmert-phytec-iot:
Generally I like the idea of drafting the manuals of unreleased BSPs in the open. We also develop them in the open and customers can test them anytime. And if they test BSPs, a draft manual is a great advantage.

Of course they have to be aware of this, so a prominent marking of draft bsps should be sufficient IMO.

source/bsp/imx8/imx8mp/pd24.1.0.rst
```diff
@@ -0,0 +1,473 @@
...
+.. include:: bootmode-switch.rsti
+
+.. include:: ../installing-os.rsti
+
+.. +---------------------------------------------------------------------------+
+.. DEVELOPMENT
+.. +---------------------------------------------------------------------------+
+
+.. _imx8mp-pd24.1.0-development:
+.. include:: /bsp/imx8/development.rsti
```
BHahn42618, line 218: You still need to rebase this to match this:
https://github.com/phytec/doc-bsp-yocto/pull/145/commits/71c580e191c3f2cf7f8c90118a3d6333f2630fc8#diff-1c7cbb8c30894749c3d2c91cfb28332d8d51bc4adb1c8e533f88b878fac3dacc

BHahn42618 (commented):
I agree with Yannic. Creating a feature branch is probably the better idea, because if we released this to public, customers would probably expect everything in the manual to work, which is not the case. That could lead to more questions for the FAE/support, which we should probably avoid.
If customers want to checkout the development version (.y manifest) they can do so based on the chapter 5.4.2 in the PD23.1.0 or head manual, where it is documented how to checkout development states of upcoming releases.

jremmert-phytec-iot:
>  [..] because if we released this to public, customers would probably expect everything in the manual to work, which is not the case. That could lead to more questions for the FAE/support, which we should probably avoid.

Yes, you have a point here. But with our existing approach, the upcoming release is already being public (HEAD Documents). Several weeks back with the PD23.1.0 imx8 release, we linked this documentation in the Phytec Website for the first time. So at this point we probably can not yet tell if there are more questions for the FAE team because of the published HEAD Documents.

The new decision here is whether we want to have a second HEAD document for a mainline release for now, as this allows a dual approach i.e. alternating mainline and NXP releases. And to make it really clear to customers, we could call it `HEAD_MAINLINE` or so instead of `PD24.1.y` unless officially released. We could  move to branches anytime, once we see support traffic going up.

If you find this too risky and expect many questions for unreleased BSPs, we should probably hide all HEAD documents in branches.

ymoog:
> If you find this too risky and expect many questions for unreleased BSPs, we should probably hide all HEAD documents in branches.

I renamed the pd24.1.y file, but kept the doc-id for now (since we also have the PD24.1.y manifest).

jremmert-phytec-iot (approved):
In Addition we could add a disclaimer note to all `HEAD` documents.

But we can do this in another PR.

EDIT: Let's wait with merging until @BHahn42618 approved, esp. regarding the FAE point.

tremmet (approved):

tremmet (approved):

BHahn42618 (approved):

PR #149, opened 2024-02-01
static: change teal color to darker Phytec color

jremmert-phytec-iot:
The darker color is already used by the documentation of modules developed in US. In addition it looks more elegant. This commit adapts the color scheme (logo background and chapter-headings).

The Chapter-Headings in HTML are not very ergonomic in terms of contrast, but there is no useful information in them anyway (Contents, Contribute) and the immediate focus falls on the headings (Yocto Reference Manuals, i.MX 8 Manuals).

Reference US Modules:
https://docs.phytec.com/latest/phycore-am62x/developingwithyocto/index.html

Old:
image

US Docs color:
image

New "teal4"
image

### DIFF
--- a/source/sphinx/static/css/phytec-theme.css
+++ b/source/sphinx/static/css/phytec-theme.css
@@ -34,12 +34,12 @@
 
 .wy-menu-vertical header,
 .wy-menu-vertical p.caption {
-    color: var(--teal3);
+    color: var(--teal4);
 
     /* Improves the appearance of uppercase text */
     letter-spacing: 0.75px;
 }
 
 .wy-side-nav-search {
-    background-color: var(--teal3);
+    background-color: var(--teal4);
 }


source/sphinx/static/css/phytec-theme.css
```diff
@@ -27,19 +27,20 @@
     --teal2: #02a6b1;
     --teal3: #16969e;
     --teal4: #006e73;
+    --phyteal: #006065;
```
mschwan-phytec, line 30: This is not ideal, as `--phyteal` does not fit in the existing color scheme, regarding its name.

Additionally, I would like to keep this theme the same as used on our website, so the branding is consistent.

I am open to style changes diverting from the default RTD theme, but they should follow the look of our website's colors and styling, if possible.

mschwan-phytec (changes requested):

source/sphinx/static/css/phytec-theme.css
jremmert-phytec-iot: Right, I agree. The closest match would be `teal4` instsead of `teal3`. I updated the description above.

mschwan-phytec (approved):

PR #155, opened 2024-02-19
bsp: imx8: installing-os: rework flash SPI NOR from SD-Card chapter

BHahn42618:
correct small mistake in mount command for sd-card. add comment that the fspi image needs to be copied to SD-Card also to the other chapter.

depends on #154 or #156


source/bsp/imx8/installing-os.rsti
```diff
@@ -471,8 +471,10 @@ Flash SPI NOR from SD Card in U-Boot on Target
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 *  Copy the SPI NOR flash U-boot image
-   imx-boot-|yocto-machinename|-fspi.bin-flash_evk_flexspi to the FAT partition
-   on the SD Card. Before reading and writing are possible, the SPI-NOR flash
+   imx-boot-|yocto-machinename|-fspi.bin-flash_evk_flexspi to the first partition
```
tremmet, line 474: I think you should not drop the information that this is the vfat partition. Maybe add this information into parentheses. Or is there any other reason why you removed it?

BHahn42618: Theoretically you could also boot from eMMC and insert an SD-Card with just one partition, that does not need to be FAT, but could also be ext4. For the command below/ chapter to work it just needs to be the first partition on the sdcard. That is why I dropped it.

tremmet: Okay, thanks.

tremmet (approved):

ymoog (approved):

jonas-rem:
@BHahn42618 can you please rebase on the current master. The linkcheck action should be ok after a rebase.

BHahn42618:
Rebased to main. Linkcheck still fails because imx93 manual contains a broken link (see #160). Added a second commit that will create the sdcarddev substitution and use the sdcard peripheral documentation from imx-common.

jonas-rem (changes requested):

BHahn42618:
Rebase to main after #160 got merged, Linkcheck passes now.

source/bsp/imx8/imx8mm/head.rst
```diff
@@ -30,6 +30,7 @@
 .. |kernel-socname| replace:: imx8mm
 .. |kernel-tag| replace:: v5.15.71_2.2.2-phy3
 .. |emmcdev| replace:: mmcblk2
+.. |sdcarddev| replace:: mmcblk1
```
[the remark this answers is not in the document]
tremmet: For i.MX95 the SD interface will be (by now) uSDCH2. As the numbering starts with 1 in HW this will likely lead to mmcblk1 in SW. 
So if nothing changes the SD-Card interface will be the same for i.MX95 and i.MX8M*. Only the eMMC will be 0 instead of 2.

jonas-rem (approved):

PR #157, opened 2024-02-27
bsp: imx8mp: mainline-head: update display section

ymoog:
For the mainline BSP, only on-board LVDS is working. There are no overlays upstream and HDMI is not supported, yet. Include only a very basic section about display support.


tremmet (approved):
Besides the one comment it looks good from my side.

jonas-rem (approved):

jonas-rem:
Linkcheck issue is known and fixed in most current main already.

PR #158, opened 2024-03-06
bsp: imx8: update links for mcore-zephyr-docs

BHahn42618:
Links for mcore-zephyr-docs substitution were broken, because the link got changed.


jonas-rem (approved):
This is a result of switching to the new `Hardware model v2` in Zephyr.

PR #160, opened 2024-03-07
linkcheck: fix broken link to Zephyr documentation

jonas-rem:
The link is currently not used in the imx93 documentation, but it causes the linkcheck to fail.


BHahn42618 (commented):
I think you can remove this line with the substitution. There is no zephyr documentation for imx93 and it currently points to the imx8mp documentation.

BHahn42618 (approved):

cstoidner (commented):
Looks fine, as zephyr is not (yet) used in i.MX 93.

PR #161, opened 2024-03-11
Rework peripherals video chapter

BHahn42618:
Rework peripherals video chapter.

use gst-play-1.0 instead of gplay-1.0, because gplay-1.0 is only in nxp-bsp. And gst-play is the same as gplay, but is also present upstream.

Change the gst-launch example with filesrc, because this was only working for h264 videos and the sync=false option causes the video to play too fast. Replace it with an option that works for all videos and will play the video at normal speed.

Remove kmssink chapter, because kmssink is not working.

Dissolve the chapter into the mainline-bsp manual and add mainline-bsp specific parameter.
Add note, that mainline BSP only supports software rendering at the moment.

### DIFF
--- a/source/bsp/peripherals/video.rsti
+++ b/source/bsp/peripherals/video.rsti
@@ -4,7 +4,8 @@ Video
 Videos with Gstreamer
 .....................
 
-The video is installed by default in the BSP:
+One example video is installed by default in the BSP at `/usr/share/qtphy/videos/`.
+Start the video playback with one of these commands:
 
 .. code-block:: console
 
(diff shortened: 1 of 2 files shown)


tremmet (approved):

source/bsp/peripherals/video.rsti
```diff
@@ -14,56 +15,10 @@ The video is installed by default in the BSP:
...
-       6 non-desktop:
-           flags: immutable range
-           values: 0 1
-           value: 0
-       4 TILE:
-           flags: immutable blob
-           blobs:
-
-           value:
+   target:~$ gst-play-1.0 /usr/share/qtphy/videos/caminandes_3_llamigos_720p_vp9.webm --videosink waylandsink
```
ymoog, line 24: or 

```bash
gst-launch-1.0 playbin uri=file:///usr/share/qtphy/videos/caminandes_3_llamigos_720p_vp9.webm
```
?

BHahn42618: in nxp-bsp playbin works, but with our current mainline BSP this does not work properly. The problem is, that with playbin I did not find a way to set a specific videosink, so it uses the autovideosink, which currently has the problem, that the video is at the wrong position on the screen. So I kept it here in the generic file for nxp-bsps and deleted it in the mainline-bsp manual.

ymoog (approved):

PR #163, opened 2024-03-25
Wip/bhahn/BSPIMX8M-3195: include chapters from imx-common in imx8 manuals

BHahn42618:
include the following chapters from imx-common:
- source/bsp/imx-common/development/format_sd-card.rsti
- source/bsp/imx-common/development/uuu.rsti
- source/bsp/imx-common/development/netboot.rsti
- source/bsp/imx-common/peripherals/i2c-bus.rsti
- source/bsp/imx-common/peripherals/network.rsti
- source/bsp/imx-common/peripherals/pin-muxing.rsti
- source/bsp/imx-common/emmc.rsti

Generalize source/bsp/imx9/peripherals/display.rsti move to imx-common and include in imx8 and imx9 @cstoidner 

Do not include bsp/imx-common/development/standalone_build.rsti, because bigger changes are needed before this can be included in imx8.

PR #164, opened 2024-03-26
BSPIMX8M-3185 Rework the network setup and network boot section to support DHCP

ymoog:
The goal is to create, in order of priority:

1. A correct BSP manual for the imx8mp mainline BSP
2. A unified manual for mainline and nxp BSPs
3. A shared/common imx manual.


BHahn42618:
Interesting. You have a much more complicated TFTP server setup. For the internal and external Tests I have done a tftp server setup on my system too. This is what my setup looks like:
`/etc/dhcp/dhcpd.conf`
```
subnet 192.168.3.0 netmask 255.255.255.0 {
range 192.168.3.11 192.168.3.250;
option broadcast-address 192.168.3.255;
option routers 192.168.3.10;
}
```
`/etc/default/isc-dhcp-server`
```
INTERFACESv4="enp53s0"
INTERFACESv6=""
```
Are your extra options really needed? Is it working without modifying the `/etc/default/isc-dhcp-server` file? Why did you choose 255.255.255.224 as a netmask and 192.168.3.31 as a broadcast address? Is that working, because on the device the ethernet interfaces are configured for netmask 255.255.255.0 and broadcast address would then be 192.168.3.255. I don't know if this could maybe cause conflicts.

source/bsp/imx8/imx8mp/mainline-head.rst
```diff
@@ -215,7 +215,192 @@ select the phyCORE-|soc| default bootsource.
...
+
+   .. code-block:: console
+
+      host:~$ sudo exportfs -ra
+
+DHCP Server setup
+~~~~~~~~~~~~~~~~~
+
+*  Create or edit the ``/etc/dhcp/dhcpd.conf`` file; Using the internal subnet
+   sample:
```
ymoog: @BHahn42618 I wrote here that the internal subnet example is being used. I modified as little as necessary.

ymoog:
>  Is that working, because on the device the ethernet interfaces are configured for netmask 255.255.255.0 and broadcast address would then be 192.168.3.255. I don't know if this could maybe cause conflicts.

The ethernet interfaces should be configured according to dhcp.
See an excerpt of the kernel log:
```
[    6.517340] IP-Config: Got DHCP answer from 192.168.3.10, my address is 192.168.3.17
[    6.525105] IP-Config: Complete:
[    6.528343]      device=eth0, hwaddr=50:2d:f4:3a:3b:e2, ipaddr=192.168.3.17, mask=255.255.255.224, gw=192.168.3.1
[    6.538616]      host=192.168.3.17, domain=internal.imx8.devboard, nis-domain=(none)
[    6.546366]      bootserver=192.168.3.10, rootserver=192.168.3.10, rootpath=
[    6.546374]      nameserver0=192.168.3.10
```

tremmet:
@ymoog I understand that you try to use an approach with minimal change. But I agree with Benjamin. We should align the DHCP configuration with the static ip settings in the BSP. And even more with the configuration for the tftp and nfs server.
They are also using a 255.255.255.0 netmask.
And 192.168.3.10 should be the router.

ymoog:
> @ymoog I understand that you try to use an approach with minimal change. But I agree with Benjamin. We should align the DHCP configuration with the static ip settings in the BSP. And even more with the configuration for the tftp and nfs server. They are also using a 255.255.255.0 netmask. And 192.168.3.10 should be the router.

Ok, I will change the netmask. However the router does not matter, only the server ip address, which is still the same as for TFTP and NFS; as can be seen in the log I provided. Thus, I would like to keep router address as is. What do you think?

tremmet:
> > @ymoog I understand that you try to use an approach with minimal change. But I agree with Benjamin. We should align the DHCP configuration with the static ip settings in the BSP. And even more with the configuration for the tftp and nfs server. They are also using a 255.255.255.0 netmask. And 192.168.3.10 should be the router.
> 
> Ok, I will change the netmask. However the router does not matter, only the server ip address, which is still the same as for TFTP and NFS; as can be seen in the log I provided. Thus, I would like to keep router address as is. What do you think?

Does does the router lead into `gw=192.168.3.1` ? I think the gateway of the target should be your host pc. Or where does this come from?

ymoog:
> > > @ymoog I understand that you try to use an approach with minimal change. But I agree with Benjamin. We should align the DHCP configuration with the static ip settings in the BSP. And even more with the configuration for the tftp and nfs server. They are also using a 255.255.255.0 netmask. And 192.168.3.10 should be the router.
> > 
> > 
> > Ok, I will change the netmask. However the router does not matter, only the server ip address, which is still the same as for TFTP and NFS; as can be seen in the log I provided. Thus, I would like to keep router address as is. What do you think?
> 
> Does does the router lead into `gw=192.168.3.1` ? I think the gateway of the target should be your host pc. Or where does this come from?

Yes. I don't believe there is such a thing as  a gateway of a physical device. However I am not sure if a gateway is needed. Maybe the gateway functionality is controlled via software on the host PC. So it acts as both a router and a client.

source/bsp/imx8/imx8mp/mainline-head.rst
```diff
@@ -215,7 +215,192 @@ select the phyCORE-|soc| default bootsource.
...
+   Be careful when creating subnets as this may interfere with the company
+   network policy. To be on the safe side, use a different network and specify
+   that via the ``interface`` configuration option.
+
+
+*  Now the DHCP-Server has to read the ``/etc/dhcp/dhcpd.conf`` file again:
+
+   .. code-block:: console
+
+         host:~# service isc-dhcp-server restart
```
BHahn42618: this did not work for me. I think the service has to be running for a restart. After the install the service was in status "failed" for me. Try this:
```suggestion
         host:~# sudo systemctl stop isc-dhcp-server
         host:~# sudo systemctl start isc-dhcp-server

**Review by BHahn42618 — CHANGES_REQUESTED** [review]

File: source/bsp/imx8/imx8mp/mainline-head.rst (hunk 2 above)
Comment (ymoog) [reply source/bsp/imx8/imx8mp/mainline-head.rst]: How about

```bash
service isc-dhcp-server --full-restart
```

jonas-rem:
@ymoog after merging #163, there are merge conflicts here. Can you rebase on the current main?

source/bsp/imx8/imx8mp/mainline-head.rst
BHahn42618: Difficult to test if this works. I got the error only when installing the dhcp-server for the first time on the host machine. When uninstalling and reinstalling it this did not work. I think I got the error because the program went into some failed state after installing it for the first time.
But I would use :~$ instead of :~# and put a sudo into the command. I don't think  that we use the :~# style anywhere and I don't think anyone is gonna get that it means that you have sudo rights.

ymoog: > Difficult to test if this works. I got the error only when installing the dhcp-server for the first time on the host machine. When uninstalling and reinstalling it this did not work. I think I got the error because the program went into some failed state after installing it for the first time. 

That is why I suggested --full-restart. It does exactly what you suggested; in one command.

> But I would use :~$ instead of :~# and put a sudo into the command. I don't think that we use the :~# style anywhere and I don't think anyone is gonna get that it means that you have sudo rights.

The $ vs # is a standard convention since this is what your prompt looks like when you are root. But I can see that this is easily missed. I'll change it to ``~$ sudo``

source/bsp/imx8/imx8mp/mainline-head.rst
ymoog: We should switch to [kea](https://ubuntu.com/server/docs/how-to-install-and-configure-isc-kea). I will take a look and modify the manual accordingly

source/bsp/imx-common/development/host_network_setup.rsti
```diff
@@ -0,0 +1,146 @@
...
+   Be careful when creating subnets as this may interfere with the company
+   network policy. To be on the safe side, use a different network and specify
+   that via the ``interfaces`` configuration option.
+
+
+*  Now the DHCP-Server has to read the ``/etc/kea/kea-dhcp4.conf`` file again:
+
+   .. code-block:: console
+
+         host:~$ sudo systemctl restart kea-dhcp4-server
```
BHahn42618, line 146: For this to work, the interface specified above needs to be active (on laptops there needs to be some device connected and powered on, so the interface does not turn off) and the interface needs to be configured to the static IP address 192.168.3.10 otherwise the dhcp server will fail. Maybe add a note for this. Otherwise it looks fine to me

ymoog: Well yes. Why do you mention it now? This is the same for all the other imx8 manuals we have. Nothing changes here. The good thing is that you are alerted to the misconfiguration where before in the other imx8 manuals you would notice this mistake later. If you would like a note to this, we should put it somewhere else where we first mention a hardcoded ip address. Do you have a suggestion for a suitable location?

tremmet (changes requested):

source/bsp/imx-common/development/host_network_setup.rsti
```diff
@@ -0,0 +1,146 @@
...
+          "subnet4": [{
+              "id": 1,
+              "next-server": "192.168.3.10",
+              "subnet": "192.168.3.0/24",
+              "pools": [
+                { "pool": "192.168.3.1 - 192.168.3.255" }
+              ]
+          }]
+        }
+      }
```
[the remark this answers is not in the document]
ymoog: Yes and no. I was not sure what to keep, so I used an example from the [doc](https://kb.isc.org/docs/kea-configuration-for-small-office-or-home-use) and reduced the content.

source/bsp/imx-common/development/host_network_setup.rsti
ymoog: I see you mean that if the network interface is not present on (host) system startup then the dhcp server startup will fail and you have to restart it once you manually set up the network interface.

In that case, I will add a note that a specific order needs to be followed when working with dhcp. Thanks for pointing this out!

tremmet (approved):

BHahn42618 (approved):

Varakala02 (approved):

PR #165, opened 2024-03-26
BSPIMX8M-3177: Wip/tremmet/i2c eeprom erase

tremmet:
Fix a documentation issue reported by the customer the when filling the EEPROM with zeros. The EEPROM data should not be overwritten. Also extend and move the EEPROM data warning to the start of the "I2C EEPROM on |som|" chapter.

@cstoidner I see that you have your own imx9 eeprom chapter. But you have basically the same issue. Should I fix it, too?

### DIFF
--- a/source/bsp/imx8/peripherals/eeprom.rsti
+++ b/source/bsp/imx8/peripherals/eeprom.rsti
@@ -23,11 +23,11 @@ execute:
 
    target:~$ dd if=/sys/class/i2c-dev/i2c-0/device/0-0051/eeprom bs=1 count=1024  | od -x
 
-To fill the whole EEPROM with zeros, use:
+To fill the 4KiB EEPROM (bus: I2C-0 addr: 0x51) with zeros leaving out the EEPROM data use:
 
 .. code-block:: console
 
-   target:~$ dd if=/dev/zero of=/sys/class/i2c-dev/i2c-0/device/0-0051/eeprom bs=4096 count=1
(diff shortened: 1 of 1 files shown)


ymoog (approved):

source/bsp/imx8/peripherals/eeprom.rsti
```diff
@@ -23,11 +23,11 @@ execute:
 
    target:~$ dd if=/sys/class/i2c-dev/i2c-0/device/0-0051/eeprom bs=1 count=1024  | od -x
 
-To fill the whole EEPROM with zeros, use:
+To fill the 4KiB EEPROM (bus: I2C-0 addr: 0x51) with zeros leaving out the EEPROM data use:
 
 .. code-block:: console
 
-   target:~$ dd if=/dev/zero of=/sys/class/i2c-dev/i2c-0/device/0-0051/eeprom bs=4096 count=1
+   target:~$ dd if=/dev/zero of=/sys/class/i2c-dev/i2c-0/device/0-0051/eeprom seek=1 bs=256 count=15
```
BHahn42618, line 30: I don't think you need the `count=15` , but in our case it does not really make a difference with or without it.

BHahn42618 (approved):

source/bsp/imx8/peripherals/eeprom.rsti
tremmet: Mhmm yes, maybe I will remove it and update the text to not be so fixated on 4kiB the eeprom. As we might also mount also different sizes.

tremmet: No it does not work. dd will complain. So I'll leave it like that.

PR #168, opened 2024-04-03
yocto: scarthgap: Add new documentation

svlstriker:
Official links to new scarthgap yocto ref-manual are
set to dev-state instead of scarthgap, as the ref-manual
is not available for scarthgap right now.

Signed-off-by: Norbert Wesp <n.wesp@phytec.de>

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -157,6 +157,14 @@ latex_documents = [
         'manual',
         False,
     ),
+    (
+        'yocto/scarthgap',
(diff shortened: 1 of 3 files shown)


jonas-rem (changes requested):
Looks good, just one remark. Could you add an entry to the `latex_documents` entry in `conf.py` so that a pdf is generated for this Yocto manual?

jonas-rem (approved):

PR #162, opened 2024-03-25
Wip/bhahn/BSPIMX8M-3179: cleanup mainline head for release

BHahn42618:
remove chapters about unsupported features:
- no SPI NOR Flash support
- no suspend to RAM support
- no support for EEPROM ID-Page
- no support for audio

PR #173, opened 2024-04-09
BSPIMX8M-3206 i.MX8MP mainline BSP manual fixups.

ymoog:
Fixes:
* yocto-mainifestname substitution has to be BSP-Yocto-Ampliphy-i.MX8MP-PD24.1.0
* 4.2.1.1. and 4.2.2.dhcp before tftp
  - add ``dhcp <image>``
* 4.2.3.1 add note that in u-boot currently only the lower usb slot works for usb sticks
* `sudo chmod a+rw /srv/tftp` to allow other users to copy to tftp directory without sudo at 4.5.1 after `sudo mkdir /srv/tftp`
   Should we give more acces rights than necessary? Best practices say no.


source/bsp/imx8/imx8mp/mainline-head.rst
```diff
@@ -211,6 +211,15 @@ select the phyCORE-|soc| default bootsource.
 .. include:: bootmode-switch.rsti
 
 .. include:: ../installing-os.rsti
+   :end-before: .. flash_emm_usb_uboot_marker
```
BHahn42618, line 214: (flash_emm_uboot marker? Change flash_emmc_usb_uboot_marker?)
This will cause merge conflicts with this PR: https://github.com/phytec/doc-bsp-yocto/pull/162/commits/d86d97b83fdeaabe89d492c9dc1f7b38d73823ad
I think we have to find the preferred solution here. Do we want to dissolve it, or do we want to include parts?

BHahn42618 (changes requested):

source/bsp/imx8/imx8mp/mainline-head.rst
ymoog: Ah, forgot about that. Let's get that merged first. The change is trivial, then.

jonas-rem (approved):

BHahn42618 (approved):

source/bsp/imx8/imx8mp/mainline-head.rst
```diff
@@ -59,7 +59,7 @@
 .. |yocto-imagename| replace:: phytec-qt6demo-image
 .. |yocto-imageext| replace:: wic
 .. |yocto-machinename| replace:: phyboard-pollux-imx8mp-3
-.. |yocto-manifestname| replace:: BSP-Yocto-i.MX8MP-PD24.1.0
+.. |yocto-manifestname| replace:: BSP-Yocto-Ampliphy-i.MX8MP-PD24.1.0
```
tremmet, line 62: You should directly update to PD24.1.1.

tremmet (changes requested):

source/bsp/imx8/imx8mp/mainline-head.rst
jonas-rem: It seems we clicked at the same time, can we change this in a follow-up?

ymoog: > You should directly update to PD24.1.1.

I disagree. We have two releases. We need 2 tags. We need two manuals. Even if they are identical.

tremmet: We do not want that people use PD24.1.0. We should not promote. It will be also extra effort in naming everything to PD24.1.0 and tag it and then do an extra step in updating everything to PD24.1.1 and make an new tag. What for?

ymoog: > What for?

ISO

tremmet: I will check if this is really required.

PR #177, opened 2024-04-16
BSPIMX8M-3204: Descibe partup issue due to new default option in resize2fs

tremmet:
Flashing a partup images created with Yocto versions mickledore or newer do not work when the systems resize2fs version is, too old (1.46.6 and older).

The newest E2fsprog versions has a new default option enabled in the filesystem, which is not compatible with older versions. 
https://e2fsprogs.sourceforge.net/e2fsprogs-release.html#1.47.0

As we support Ubuntu 22.04 as host system we do see the issue there.

Make flashing an SD-Card from host with bmap as first option in the documentation and add a note in the partup section.
As the same issue appears when using gparted and a note there, too.

### DIFF
--- a/source/bsp/getting-started.rsti
+++ b/source/bsp/getting-started.rsti
@@ -97,35 +97,10 @@ Unmount all those partitions, e.g.:
 Now, the SD card is ready to be flashed with an image, using either ``partup``,
 ``dd`` or ``bmap-tools``.
 
-Using partup
-............
-
-Writing to an SD card with partup is done in a single command:
-
-.. code-block:: console
-   :substitutions:
-
-   host:~$ sudo partup install |yocto-imagename|-|yocto-machinename|.partup /dev/<your_device>
-
-Make sure to replace <your_device> with your actual device name found previously.
-
-Further usage of partup is explained at its `official documentation website
-<https://partup.readthedocs.io/en/latest/>`__.
-
-.. note::
-   *partup* has the advantage of allowing to clear specific raw areas in the
-   MMC user area, which is used in our provided partup packages to erase any
-   existing U-Boot environments. This is a known issue *bmaptool* does not
-   solve, as mentioned below.
-
-   Another key advantage of partup over other flashing tools is that it allows
-   configuring MMC specific parts, like writing to eMMC boot partitions, without
-   the need to call multiple other commands when writing.
-
 Using bmap-tools
 ................
 
-An alternative and also fast way to prepare an SD card is using
+One way to prepare an SD card is using
 `bmap-tools <https://github.com/yoctoproject/bmaptool>`_. Yocto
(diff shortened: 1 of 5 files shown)


source/bsp/getting-started.rsti
```diff
@@ -153,6 +128,37 @@ to skip.
...
+
+   host:~$ sudo partup install |yocto-imagename|-|yocto-machinename|.partup /dev/<your_device>
+
+Make sure to replace <your_device> with your actual device name found previously.
+
+Further usage of partup is explained at its `official documentation website
+<https://partup.readthedocs.io/en/latest/>`__.
+
+.. warning::
+   *partup* running on host systems which are using resize2fs version 1.46.6 and older
```
mschwan-phytec, line 147: The sentence is a little weird to read, IMHO. I think it can be easily fixed by removing the following part at the beginning:
```suggestion
   Host systems which are using resize2fs version 1.46.6 and older
```

source/bsp/getting-started.rsti
```diff
@@ -153,6 +128,37 @@ to skip.
...
+   (e.g. Ubuntu 22.04) are not able to write a partup image created with Yocto Mickledore
+   or newer to SD-Card. This is due to a new default option in resize2fs which causes a
```
mschwan-phytec, line 149: "a**n** incompatibility"
```suggestion
   or newer to SD-Card. This is due to a new default option in resize2fs which causes an
```

source/bsp/getting-started.rsti
```diff
@@ -153,6 +128,37 @@ to skip.
...
+   host:~$ sudo partup install |yocto-imagename|-|yocto-machinename|.partup /dev/<your_device>
+
+Make sure to replace <your_device> with your actual device name found previously.
+
+Further usage of partup is explained at its `official documentation website
+<https://partup.readthedocs.io/en/latest/>`__.
+
+.. warning::
+   *partup* running on host systems which are using resize2fs version 1.46.6 and older
+   (e.g. Ubuntu 22.04) are not able to write a partup image created with Yocto Mickledore
```
mschwan-phytec, line 148: `s/partup image/partup package/g`
We should be consistent with the naming of "partup packages".
```suggestion
   (e.g. Ubuntu 22.04) are not able to write partup packages created with Yocto Mickledore
```

mschwan-phytec (changes requested):

ymoog (approved):

BHahn42618 (approved):

source/bsp/getting-started.rsti
```diff
@@ -153,6 +128,37 @@ to skip.
...
+   or newer to SD-Card. This is due to a new default option in resize2fs which causes a
+   incompatibility.
```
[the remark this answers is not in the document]
mschwan-phytec: I also noticed that. :confused:

mschwan-phytec (approved):
Thanks for the changes!

PR #183, opened 2024-06-04
yocto: mickledore: Add new compatible BSP

svlstriker:
Also updated phyLinux init walkthrough.


cstoidner (approved):
This all looks good. Reviewed as correct and accepted.

PR #186, opened 2024-06-10
BSPIMX8M-2758: Update imx8 mainline HEAD for flashing wic.xz images

ymoog:
For the next mainline release, imx8 will switch to using compressed (wic.xz) OS images.

To this extent, the mainline-head imx8mp document needs to be changed.

- When flashing an image using dd, image needs to be decompressed first
- When flashing an image through the bootloader shell, image need to be decompressed first.
- Image filename extension will be updated.
- update list of files generated through the build system
- update misc. hardcoded .wic image names where necessary


source/bsp/imx8/imx8mp/mainline-head.rst
```diff
@@ -480,14 +303,14 @@ You can update the eMMC from your target.
...
-the network (then uncompress it, if necessary) to the eMMC of the target with a
+Take a compressed or uncompressed image with accompanying block map on the host
+and send it with ssh through the network to the eMMC of the target with a
 one-line command:
 
 .. code-block:: console
    :substitutions:
 
-   target:~$ ssh <USER>@192.168.3.10 "dd if=<path_to_file>/|yocto-imagename|-|yocto-machinename|.|yocto-imageext|" | dd of=/dev/mmcblk2 conv=fsync
+   target:~$ scp <USER>@192.168.3.10:/srv/tftp/|yocto-imagename|-|yocto-machinename|.wic.* /tmp && bmaptool copy /tmp/|yocto-imagename|-|yocto-machinename|.|yocto-imageext| /dev/mmcblk2
```
tremmet, line 313: I would use here `|yocto-imageext|` instead of `wic.*`. This prevents any confusion.

What about the "rootfs" in the image name here?

source/bsp/imx8/installing-os.rsti
```diff
@@ -112,14 +118,14 @@ You can update the eMMC from your target.
...
-the network (then uncompress it, if necessary) to the eMMC of the target with a
+Take a compressed or uncompressed image with accompanying block map on the host
+and send it with ssh through the network to the eMMC of the target with a
 one-line command:
 
 .. code-block:: console
    :substitutions:
 
-   target:~$ ssh <USER>@192.168.3.10 "dd if=<path_to_file>/|yocto-imagename|-|yocto-machinename|.wic" | dd of=/dev/mmcblk2 conv=fsync
+   target:~$ scp <USER>@192.168.3.10:/srv/tftp/|yocto-imagename|-|yocto-machinename|.wic.* /tmp && bmaptool copy /tmp/|yocto-imagename|-|yocto-machinename|.|yocto-imageext| /dev/mmcblk2
```
tremmet, line 128: What about `rootfs` in the image name here?

source/bsp/imx8/installing-os.rsti
```diff
@@ -348,24 +338,8 @@ image saved on the SD card.
       target:~$ ls
       |yocto-imagename|-|yocto-machinename|.partup
       |yocto-imagename|-|yocto-machinename|.wic
+      |yocto-imagename|-|yocto-machinename|.wic.bmap
```
tremmet, line 341: rootfs in the image name?

tremmet (changes requested):

source/bsp/imx8/imx8mp/mainline-head.rst
ymoog: I used wic.* because the bmap should be copied along with the image.

> What about the "rootfs" in the image name here?

It is not clear if we will keep rootfs in the image name (see internal discussions). As a result, I left it as is.

tremmet: > I used wic.* because the bmap should be copied along with the image.

Allright.
> 
> > What about the "rootfs" in the image name here?
> 
> It is not clear if we will keep rootfs in the image name (see internal discussions). As a result, I left it as is.

But it is only important if it makes into the BSP or? So you want to wait and update later? With this state we have to update it for sure. Whatever the outcome is.

ymoog: Yes. When we start working on the next Vendor BSP, we will need to either add rootfs to this doc or remove it in our Yocto layer.

Varakala02 (approved):

source/bsp/imx8/imx8mp/mainline-head.rst
ymoog: > > What about the "rootfs" in the image name here?
> 
> It is not clear if we will keep rootfs in the image name (see internal discussions). As a result, I left it as is.

Apologies, rootfs should be included in this file, because it is necessary for the next PD24.1.2 release. In the other occasions for the non-mainline version not including rootfs remains correct.

source/bsp/imx8/installing-os.rsti
tremmet: Okay we will only add the rootfs in mainline-head for now.

source/bsp/imx8/installing-os.rsti
tremmet: Same here. Change only for mainline head.

tremmet (dismissed):

mschwan-phytec (changes requested):

source/bsp/imx8/imx8mp/mainline-head.rst
```diff
@@ -612,33 +437,16 @@ need a complete image saved on the USB stick and a bootable WIC image. (e.g.
...
-      mmcblk1p2
-      mmcblk2
-      mmcblk2boot0
-      mmcblk2boot1
-      mmcblk2p1
-      mmcblk2p2
-      mmcblk2rpmb
-
-*  The eMMC device can be recognized by the fact that it contains two boot
-   partitions: (mmcblk2boot0; mmcblk2boot1)
```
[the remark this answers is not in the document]
ymoog: In short: I deemed this unnecessary and inconsistent for the flashing explanation. 

When we hardcode the device in the next section just after this one, what use does this ``ls /dev`` have? It may be useful for general explanation of how block devices are displayed in Linux, but I found it rather confusing since nothing said here has any impact on the instructions to be executed.

source/bsp/imx8/imx8mp/mainline-head.rst
```diff
@@ -721,24 +529,8 @@ image saved on the SD card.
...
-      mmcblk1p2
-      mmcblk2
-      mmcblk2boot0
-      mmcblk2boot1
-      mmcblk2p1
-      mmcblk2p2
-      mmcblk2rpmb
-
-*  The eMMC device can be recognized by the fact that it contains two boot
-   partitions: (mmcblk2\ **boot0**; mmcblk2\ **boot1**)
```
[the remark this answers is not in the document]
ymoog: Same reasoning as above.

source/bsp/imx8/installing-os.rsti
```diff
@@ -1,6 +1,12 @@
 Flash eMMC
 ----------
 
+For consistency, it is assumed that a TFTP server is configured; More
+importantly, all generated images, as listed above, are copied to the default
+/srv/tftp directory. If you do not have this set up, you need to adjust the
+paths that point to the images being used in the instructions. For instructions
+on how to set up the TFTP server and directory, see |ref-setup-network-host|.
+
```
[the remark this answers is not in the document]
ymoog: Users may choose their preferred path:

> If you do not have this set up, you need to adjust the
paths that point to the images being used in the instructions.

---
> What let to this decision?

The fact, that more commands may be copied as is. Fewer modifications to the supplied commands are necessary this way. Also, consistency.

Let's look at an example:

```diff
-   host:~$ dd if=|yocto-imagename|-|yocto-machinename|.|yocto-imageext| status=progress | ssh root@192.168.3.11 "dd of=/dev/mmcblk2 conv=fsync"
+   host:~$ scp /srv/tftp/|yocto-imagename|-|yocto-machinename|.|yocto-imageext| root@192.168.3.11:/tmp && ssh root@192.168.3.11 "bmaptool copy /tmp/|yocto-imagename|-|yocto-machinename|.|yocto-imageext| /dev/mmcblk2"
```

Before, you needed to change directory to where the image is located. 
Now, copy command as is.


To expand further: We have a mixture of relative path, i.e. ``cp <image> /someplace/else`` and placeholder ``cp <path/to/image>/image /someplace/else``. 
One could make the argument to use relative paths for every command, however I found using an absolute path more robust (no need to pay attention the cwd). 
Also, when/if using netboot/TFTP, images are already at a predefined location. This location may easily be reused.

source/bsp/imx8/installing-os.rsti
```diff
@@ -136,16 +142,17 @@ Show your available image files on the host:
 .. code-block:: console
    :substitutions:
 
-   host:~$ ls
+   host:~$ /srv/tftp
    |yocto-imagename|-|yocto-machinename|.wic
+   |yocto-imagename|-|yocto-machinename|.wic.bmap
```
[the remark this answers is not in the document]
ymoog: > Did you mean `ls /srv/tftp`? Anyhow, please read my other comment above on using this path, first.

yep, messed that up.

source/bsp/imx8/imx8mp/mainline-head.rst
```diff
@@ -480,14 +303,14 @@ You can update the eMMC from your target.
 
    A working network is necessary! |ref-setup-network-host|
 
-Take a compressed or uncompressed image on the host and send it with ssh through
-the network (then uncompress it, if necessary) to the eMMC of the target with a
+Take a compressed or uncompressed image with accompanying block map on the host
+and send it with ssh through the network to the eMMC of the target with a
```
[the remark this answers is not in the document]
ymoog: Strictly speaking, it is not clear whether the image had been compressed before, so using uncompressed would be fine.

In the context of the release, we only supply compressed images. So, unless an umcopressed image had been generated at another point, it is safe to assume the image referred to here had to have been compressed.

I'll allow it.

source/bsp/imx8/imx8mp/mainline-head.rst
```diff
@@ -504,16 +327,17 @@ Show your available image files on the host:
...
 
-Send the image with the ``dd`` command combined with ssh through the network
+Send the image with the ``bmaptool`` command combined with ssh through the network
 to the eMMC of your device:
 
 .. code-block:: console
    :substitutions:
 
-   host:~$ dd if=|yocto-imagename|-|yocto-machinename|.|yocto-imageext| status=progress | ssh root@192.168.3.11 "dd of=/dev/mmcblk2 conv=fsync"
+   host:~$ scp /srv/tftp/|yocto-imagename|-|yocto-machinename|.|yocto-imageext| root@192.168.3.11:/tmp && ssh root@192.168.3.11 "bmaptool copy /tmp/|yocto-imagename|-|yocto-machinename|.|yocto-imageext| /dev/mmcblk2"
```
tremmet, line 340:
```suggestion
   host:~$ scp /srv/tftp/|yocto-imagename|-|yocto-machinename|.rootfs.wic.* root@192.168.3.11:/tmp && ssh root@192.168.3.11 "bmaptool copy /tmp/|yocto-imagename|-|yocto-machinename|.|yocto-imageext| /dev/mmcblk2"
```

tremmet (changes requested):

source/bsp/getting-started.rsti
```diff
@@ -111,12 +111,12 @@ Debian-based systems install it by issuing:
 
    host:~$ sudo apt install bmap-tools
 
-Flash a WIC image to SD card by calling:
+Flash an uncompressed WIC image to SD card by calling:
```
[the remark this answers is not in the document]
ymoog: I did that because I did not want to alter the manual where not necessary. However, since bmaptool does not care whether the image is compressed or not we can leave it as is. Good point, thanks!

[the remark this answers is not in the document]
ymoog: I used the same convention as for the optional ?(.rootfs) to support .wic and .wic.xz.

tremmet (dismissed):

tremmet (approved):

Varakala02 (approved):

jonas-rem (approved):
I did not found showstoppers in the unresolved conversations, so I will merge the PR. We can always open a follow-up PR to continue to discuss those issues and find optimal solutions.

PR #187, opened 2024-06-10
Add config file for rstcheck

ymoog:
I am using [rstcheck](https://github.com/rstcheck/rstcheck) as a static analysis tool. Add an initial config file for the analyzer to ignore substitutions it cannot handle. These are enhanced/custom ones introduced by sphinx extensions.

### DIFF
--- a/.rstcheck.cfg
+++ b/.rstcheck.cfg
@@ -1,0 +1,15 @@
+[rstcheck]
+ignore_substitutions=
+   yocto-bootenv-link,
+   ref-bootswitch,
+   ref-bsp-images,
+   ref-debugusbconnector,
+   ref-dt,
+   ref-getting-started,
+   ref-network,
+   ref-setup-network-host,
+   ref-usb-otg,
+   ref-disable-emmc-part,
+   ref-serial,
+   ref-jp3,
+   ref-jp4


mschwan-phytec (changes requested):
If this config is not used in any GitHub Action, I'd would kindly ask *not* to include this. I'd like to avoid adding "random" configs for userspace tools. It is up to the user to decide which tools they edit this source code with and need. Exaggerating, we cannot store all config files for all the third-party programs out there. I'd like to avoid tracking unused (regarding the source code) files in this repository. IMHO, you are better off saving this config e.g. in your personal dotfiles. I hope, I don't come across as mean now. :grimacing: Kind regards!

ymoog:
> I'd like to avoid tracking unused (regarding the source code) files in this repository. 

Makes sense to me.

>IMHO, you are better off saving this config e.g. in your personal dotfiles. I hope, I don't come across as mean now. 😬 Kind regards!

In the user configs goes general stuff. What I add is project specific and thus should go in the project directory.
 
If it is unacceptable to you to have this config tracked, my suggestion is to have it added to gitignore instead.

mschwan-phytec:
> In the user configs goes general stuff. What I add is project specific and thus should go in the project directory.

True. I am not familiar with rstcheck specifically, but maybe it allows for storing a global config applying to all projects. That was my thinking.

> If it is unacceptable to you to have this config tracked, my suggestion is to have it added to gitignore instead.

That's okay for me, although the same applies here: We do not want to add all possible files to ignore in `.gitignore`. But I think we can keep it clear and a few lines there won't hurt. :+1:

mschwan-phytec:
I'm closing this then, as I think we came to a common ground.

@ymoog Please open a separate PR, if you want the rstcheck config added to `.gitignore`.

PR #203, opened 2024-07-09
GH actions: add status check with documentation preview

jonas-rem:
GitHub pages can host documentation previews for each pull request. This can be helpful for reviewing and getting an overview of changes quickly.

For testing check the passed checks:
`Check External Links / Click on details for a Doc Preview (pull_request)`

PR #204, opened 2024-07-09
PDF: add download links to all manuals

jonas-rem:
This PR adds the option to download a manual as a pdf file.

Currently there are two issues with the pdf files:

1. The release in the pdf files is the documentation release (e.g. `imx8mp-pd24.1.2-2-gb7c1ba7`). This is confusing for customers that e.g. view an imx93 manual with that release string. Renaming the release string from `Release` to `Documentation Build` may reduce confusions.
2. When providing pdf documents for head versions, we should probably mark them clearly to indicate that those are not final releases.

**Edit:**
To 1.: The release version is no longer shown on the title page, but on the left side of the header as `Documentation rev.`:.
To 2: The right side of the header now shows the release number or `DRAFT` for head manuals.

**Note:** As this PR is based on #203, you can review the pdf download feature with the documentation preview down in the checks.

~**To all Reviewers:** Do you see a benefit to our customers by providing pdf files in this way?~


ymoog (approved):

s-hemer (dismissed):

s-hemer:
The version string is indeed confusing. Although, I do not have a direct idea how to solve it, I am pretty sure that both issues can be solve in one step.

jonas-rem:
I addressed the remaining issues (see updated description) and linked pdfs for all documents (en, zh_CN) are available now.

ymoog (approved):

PR #214, opened 2024-09-02
imx-common: standalone_build_kernel: Fix make parameter

tremmet:
Make sure that the nproc command is actually called and not that a variable is passed. As if ${nproc} is empty it leads to unlimited number of jobs being created and can make the system unresponsive.


jremmet (approved):

ymoog (approved):
I see :)

mschwan-phytec (approved):

PR #218, opened 2024-09-18
bsp: imx-common: standalone_build_kernel: Fix indentation

s-hemer:
The code block wasn't correctly formatted. Fixes 9badff3380f40

### DIFF
--- a/source/bsp/imx-common/development/standalone_build_kernel.rsti
+++ b/source/bsp/imx-common/development/standalone_build_kernel.rsti
@@ -38,8 +38,8 @@ Build Kernel
 
 *  Install kernel modules to e.g. NFS directory:
 
-  .. code-block:: console
-     :substitutions:
+   .. code-block:: console
+      :substitutions:
 
       host:~/|kernel-repo-name|$ make INSTALL_MOD_PATH=/home/<user>/<rootfspath> modules_install
 


s-hemer:
BTW this code block seems to be somehow out of context: bullet point before is "building kernel", following is about where to find the results. Imho installing kernel modules is an more advanced task and should go to the end of the list (behind re-building DTs)?!

tremmet:
> BTW this code block seems to be somehow out of context: bullet point before is "building kernel", following is about where to find the results. Imho installing kernel modules is an more advanced task and should go to the end of the list (behind re-building DTs)?! Potentially, this is also a duplicate info as it is part of the following section, so completely remove here?

Feel free to create a pull request or write a ticket in our jira project.

tremmet:
@s-hemer don't forget to add reviewers to your pull request :)

tremmet (approved):

source/bsp/imx-common/development/standalone_build_kernel.rsti
```diff
@@ -39,7 +39,7 @@ Build Kernel
 *  Install kernel modules to e.g. NFS directory:
 
   .. code-block:: console
```
BHahn42618, line 41: This line is also missing one space.
```suggestion
   .. code-block:: console
```

BHahn42618 (changes requested):

source/bsp/imx-common/development/standalone_build_kernel.rsti
s-hemer: Fixed

BHahn42618 (approved):

tremmet (approved):

PR #219, opened 2024-09-19
initial translation version

annie-pixel:


jonas-rem (changes requested):
Can you rebase on the current branch `WIP/j.remmert@phytec.de/internationalization`? I was squashing the last commit with the example changes of the po file into the previous commit.

In addition I saw that in `bsp.po` there are Windows-Style CR/LF. Can you check and convert them to Unix style? I don't know why it shows only in this file, usually git should automatically convert them.

Edit: Can you also squash your last two commits into the first one, so that you only have one commit?

annie-pixel:
create a new pull request

PR #223, opened 2024-09-27
Wip/bhahn/BSPIMX8M-3122: add back uboot build with fixed ram freq

BHahn42618:
The imx8 uboot-standalone build is not much different from the
imx-common uboot-standalone build. The only thing done there are some
improvements in the structuring of the standalone build chapter. Apply
this restructuring to imx-common and old PD22.1.1 manuals, so we can
continue to include imx-common standalone build chapter for new imx8mp
NXP PD24.1.0 release.
   
As imx93 also uses imx-common, it will also get the new structure. For
imx93 there will not be any functional changes except for the new
structuring.
    
Some minor fixes were applied here. For example also use the
ref-build-uboot link in imx-common. And include the build with fixed
ram size also for imx8mm pd22.1.1 as it was missing.
    
Move the imx8mp specific build with fixed ramsize and timings to its own
file, so it can be included in the newer imx8mp releases.

add back uboot build with fixed ram freq for newer imx8mp releases


source/bsp/imx8/imx8mp/development/uboot-standalone-fixed-ramsize.yaml
ymoog, line 1: With this change, it makes sense to rename the file.
Call it uboot-ram-configuration or something that's to your liking.

ymoog (changes requested):

source/bsp/imx8/imx8mp/development/uboot-standalone-fixed-ramsize.yaml
BHahn42618: Should be `.rsti` not `.yaml`. I will fix that. Interesting that it worked with .yaml and nothing complained, but what is wrong with `uboot-standalone-fixed-ramsize.rsti`? (Before it was just `uboot-standalone.rsti`, so I already did some renaming here)

ymoog: >  what is wrong with uboot-standalone-fixed-ramsize.rsti?

You add RAM frequency and at that point the name is no longer accurate.

BHahn42618: I renamed to uboot-standalone-fixed-ram-config.rsti

tremmet (changes requested):

tremmet (approved):

BHahn42618:
Yannic is on vacation

ymoog (approved):

PR #224, opened 2024-09-30
BSPIMX8M-2641: bsp: imx8: installing-os: Update default eMMC flashing

Varakala02:
Set the default or initial flashing of eMMC via kernel. The U-Boot way of eMMC flashing is slow and image size is limited by RAM space. So, updated with flashing eMMC via kernel sections first before U-Boot.

Also Updated for SPI NOR Flash sections.

### DIFF
--- a/source/bsp/imx8/installing-os.rsti
+++ b/source/bsp/imx8/installing-os.rsti
@@ -404,38 +404,6 @@ populated on the phyCORE-|soc|.
 
    A working network is necessary! |ref-setup-network-host|
 
-Flash SPI NOR from Network in U-Boot on Target
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-Similar to updating the eMMC over a network, be sure to set up the development
-host correctly. The IP needs to be set to 192.168.3.10, the netmask to
-255.255.255.0, and a TFTP server needs to be available. Before reading and
-writing is possible, the SPI NOR flash needs to be probed:
-
-.. code-block::
-
-   u-boot=> sf probe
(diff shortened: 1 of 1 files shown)


source/bsp/imx8/installing-os.rsti
```diff
@@ -34,6 +34,51 @@ Therefore, it is possible to flash the **WIC image** (``<name>.wic``) from
 the Yocto build system directly to the eMMC. The image contains the
 bootloader, kernel, device tree, device tree overlays, and root file system.
 
+Flash eMMC via Network in Linux on Target
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
BHahn42618, line 38: I would probably prefer `Flash eMMC via Network in Linux on Host` before this one (`Flash eMMC via Network in Linux on Target`). But other than that it looks fine to me.

BHahn42618 (commented):
I would prob

source/bsp/imx8/installing-os.rsti
Varakala02: Yes, I agree, thanks, I will update it.

tremmet:
Even though this might not be part of the ticket and pull request but can you also swap SPI NOR flashing chapters while your on it?

Varakala02:
> Even though this might not be part of the ticket and pull request but can you also swap SPI NOR flashing chapters while your on it?

Sure, I will do it.

tremmet (approved):

BHahn42618 (approved):

BHahn42618 (approved):

PR #225, opened 2024-10-08
peripherals: emmc: Remove resize2fs section

tremmet:
With scarthgap imx8mp-nxp the mentioned resize2fs command is not working as noted in the example. It is just doing nothing.

root@phyboard-pollux-imx8mp-3:~# resize2fs /dev/mmcblk2p2 resize2fs 1.47.0 (5-Feb-2023)
The filesystem is already 422992 (4k) blocks long.  Nothing to do!

The previous described parted command is doing its job if a partition is mounted or not. So we can just dump the whole resize2fs section.

### DIFF
--- a/source/bsp/peripherals/emmc.rsti
+++ b/source/bsp/peripherals/emmc.rsti
@@ -238,20 +238,6 @@ Increasing the filesystem size can be done while it is mounted.  But you can
 also boot the board from an SD card and then resize the file system on the eMMC
 partition while it is not mounted.
 
-*  Resize the filesystem to a new partition size:
-
-   .. code-block:: console
-      :substitutions:
-
-      target:~$ resize2fs /dev/|emmcdev|p2
-      resize2fs 1.46.1 (9-Feb-2021)
-      Filesystem at /dev/|emmcdev|p2 is mounted on /; on-line resizing required
-      [ 131.609512] EXT4-fs (|emmcdev|p2): resizing filesystem
-      from 454136 to 7367680 blocks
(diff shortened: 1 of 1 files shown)


mschwan-phytec:
This seems wrong. `parted resizepart` only changes the partition's end position in the partition table, but not the filesystem: https://www.gnu.org/software/parted/manual/html_node/resizepart.html

Did you check, whether your partition actually was resized after the `parted resizepart` command? When booting from the same system you are resizing the rootfs from, then you have to reboot in order for the kernel to pick up the new partition sizes.

tremmet:
> This seems wrong. `parted resizepart` only changes the partition's end position in the partition table, but not the filesystem: https://www.gnu.org/software/parted/manual/html_node/resizepart.html That is also why `parted resizepart` works for both mounted and not mounted partitions, as it does not change the filesystem in use.
> 
> Did you check, whether your partition actually was resized after the `parted resizepart` command? When booting from the same system you are resizing the rootfs from, then you have to reboot in order for the kernel to pick up the new partition sizes.

Okay I understand. I guess I was rushing too fast. I will retest this an probably this needs just a little rewording to make it more clear...

mschwan-phytec:
Yes, I think some rewording would help here, to make this clear.

mschwan-phytec:
@tremmet May you want to open a new PR with some rewording, if necessary, and we close this one?

tremmet:
> @tremmet May you want to open a new PR with some rewording, if necessary, and we close this one?

If you prefer we can close this one and I will open a new one.

mschwan-phytec:
Okay, closing this one then.

PR #229, opened 2024-10-15
initial chinese translation

annie-pixel:
initial chinese translation for bsp, yocto and contributing links


jonas-rem (approved):
Looks good to me!

There was one problem that we discussed offline:
When editing translations with `Poedit`, the line-breaks are handled differently than `Sphinx` is doing. Those small differences (e.g. whitespace at the beginning of the line instead of at the end) lead to a large diff when the translation message files are rebuild after changes.

For that reason we decided to disable linebreaks in Sphinx as well as in Poedit.

I will create a follow-up PR with:
- updated tox build setting so that `Sphinx` does not wrap lines
- updated Contributing guide where we mention the required settings in `Poedit`.

PR #230, opened 2024-10-15
BSPIMX8M-3496 adapt dt-overlays chapter to external boot script

ymoog:
Handling of device tree overlays changed due to introduction of loading and executing an external boot script.
As the boot script can be used for all PHYTEC products (using U-Boot) make a generic include file for overlay handling.
The new file is mostly a copy of the existing imx8 one, but does not contain any product specific info (examples include a specific expansion board, though).

This PR introduces a generic dt-overlays file, that may be included by any product family.
In addition the imx8mp includes this newly generated file as it is the only product making use of the U-Boot standard boot bootflow.

### DIFF
--- a/source/bsp/dt-overlays.rsti
+++ b/source/bsp/dt-overlays.rsti
@@ -1,0 +1,138 @@
+The usage of overlays can be configured during runtime in Linux or U-Boot.
+Overlays are applied during the boot process in the bootloader after the
+boot command is called and before the kernel is loaded. The next sections
+explain the configuration in more detail.
+
+Set ``${overlays}`` variable
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``${overlays}`` U-Boot environment variable contains a number-sign (#)
+separated list of overlays that will be applied during boot. The overlays
+listed in the overlays variable must be included in the FIT image. Overlays set
+in the $KERNEL_DEVICETREE Yocto machine variable will automatically be added to
+the FIT image.
+
+The ``${overlays}`` variable can either be set directly in the U-Boot
+environment or can be part of the external ``bootenv.txt`` environment file.
+When desired to use the overlays variable as set manually in the U-Boot
+environment, disable bootenv by setting ``env set no_bootenv 1`` as the overlays
+variable may be overwritten during the execution of the boot script.
+By default, the ``${overlays}`` variable comes from the external ``bootenv.txt``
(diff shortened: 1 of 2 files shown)


source/bsp/dt-overlays.rsti
```diff
@@ -0,0 +1,148 @@
...
+   u-boot=> env delete overlays
+   u-boot=> env set no_bootenv 1
+
+
+If desired to use the bootenv.txt file for setting U-Boot variables other than
+overlays and having overlays disabled, remove the overlays definition line from
+the bootenv.txt file instead of setting no_bootenv.
+
+Extension Command
+~~~~~~~~~~~~~~~~~
```
BHahn42618, line 71: The extension command is used for legacy boot way. For fitimage boot, there is a function called `phytec_set_fit_extensions` in the board code, which sets the variable `fit_extensions` in the uboot environment. The same overlays should appear there as in the extension command, but they are formatted like in the overlays variable with `conf-overlayname1#conf-onverlayname2`. The `setenv no_extensions 1` command will also work for fitimage boot though.

BHahn42618 (changes requested):

source/bsp/dt-overlays.rsti
ymoog: The function is specific for imx8mp an as such not fit for inclusion here. Should we leave that section out for now?

BHahn42618: I think we should still describe, that some overlays that can be detected via eeprom get loaded automatically and that the loading mechanism can be disabled by setting no_extensions.
You are correct however. The mechanism how we did it is imx8 specific and should not be in this chapter.

source/bsp/dt-overlays.rsti
```diff
@@ -0,0 +1,138 @@
...
+variable may be overwritten during the execution of the boot script.
+By default, the ``${overlays}`` variable comes from the external ``bootenv.txt``
+environment file which is located in the boot partition.
+You can read and write the file on booted target from linux:
+
+.. code-block:: console
+   :substitutions:
+
+   target:~$ cat /boot/bootenv.txt
+   overlays=conf-|dt-carrierboard|-peb-eval-01.dtbo#|dtbo-peb-av-10|
```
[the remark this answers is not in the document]
ymoog: Yes. Any name/overlay will do as it is an example.

BHahn42618 (approved):

PR #231, opened 2024-10-16
zh_CN: hide translation until final review

jonas-rem:


annie-pixel (approved):
okay,block chinese doc first

PR #242, opened 2024-10-31
imx8mp: head: Remove BSP extensions chapter

tremmet:
This chapter contains so far only the description in how to build Chromium but this is broken now for the second release. As we do not actively support it, remove the chapter from the i.MX8MP documentation.


tremmet:
External Link works. But takes quite a while to load...

ymoog (approved):

BHahn42618 (approved):

jonas-rem:
> External Link works. But takes quite a while to load...

Yes, it's a known issue with our internal servers. We can ignore the check for now.

jonas-rem (approved):

PR #243, opened 2024-10-31
bsp: imx93: add pdf build, link to html documentation

jonas-rem:
As described in #228, this is a follow-up PR to enable and link the pdf version of the manual.


pfiser:
@jonas-rem Thanks for taking care of this. LGTM.

pfiser (approved):

PR #244, opened 2024-10-31
seo: fix sitemap, remove alternate links

jonas-rem:
When adding a language prefix, this would also affect the english version, which currently does not have a language code as it is the default. Therefore remove alternative links and only add the basic links without language code to the sitemap.

Fix to #236  after checking google search console.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -78,8 +78,11 @@ html_extra_path = ['sphinx/static/robots.txt',
                    'sphinx/static/googledd12f2e41658d61e.html']
 
 # Link scheme
-sitemap_url_scheme = "{lang}{link}"
-sitemap_locales = ['zh_CN']
+sitemap_url_scheme = "{link}"
+# When adding a lang prefix, this would also affect the english version, which
+# currently does not have a language prefix as it is the default. Therefore, we
+# currently only support a sitemap without alternative links.
+sitemap_locales = [None]
 
 html_theme_options = {
     'logo_only': False,


source/conf.py
```diff
@@ -78,8 +78,11 @@
                    'sphinx/static/googledd12f2e41658d61e.html']
 
 # Link scheme
-sitemap_url_scheme = "{lang}{link}"
-sitemap_locales = ['zh_CN']
+sitemap_url_scheme = "{link}"
+# When adding a lang prefix, this would also affect the english version, which
```
ymoog, line 82: Can you elaborate in what way this would have an effect. E.g. is this positive or negative and why?

jonas-rem: Sure. This change has an effect on the generated sitemap.xml file. The previous schema generated links by adding a language prefix (en|zh_CN) before the link, e.g. https://phytec.github.io/doc-bsp-yocto/en/bsp/imx8/imx8mn/head.html.

See https://phytec.github.io/doc-bsp-yocto/sitemap.xml

The current documentation structure, however, assumes english as the default language that does not use a language prefix. Other language like zh_CN still have the prefix. The sphinx plugin that generates the sitemap does not support specifying different structures, so you have to either add the prefix everywhere or nowhere.

To fix this I see two options:
1. Only use the default language links (en) which does not need a prefix. Drop other languages as alternative links from the sitemap.
2. Change the deault structure so that it also uses a language prefix. This means all documentation links will change and the links become longer.

This PR implements option 1. Other languages are not listed in the sitemap.xml anymore.

In general I hope to improve the listing of the documentation in e.g. google search by providing a functioning sitemap file.

ymoog: I don't quite understand how the problem relates to the sitemap.xml file. What am I supposed to see in the sample you gave?

jonas-rem: This PR changes how the Sphinx plugin `sphinx_sitemap` generates the sitemap. The sitemap is automatically generated accordingly when invoking the regular Sphinx html build.

The documentation of the plugin explains how the URL Scheme can be configured https://sphinx-sitemap.readthedocs.io/en/latest/advanced-configuration.html#customizing-the-url-scheme

By removing `{lang}` from `sitemap_url_scheme` the language prefixes are not added when generating sitemap.xml.

Setting `sitemap_locales = [None]` removes alternative language links (zh_CN in this case).

ymoog (approved):

PR #252, opened 2024-11-13
bsp: imx8mp: Fix remoteproc overlay substitution variable

Varakala02:
Update the remoteproc substition variable "dtbo-rpmsg" with the prefix "conf-" for head.rst and pd24.1.0_nxp.rst.

### DIFF
--- a/source/bsp/dt-overlays.rsti
+++ b/source/bsp/dt-overlays.rsti
@@ -39,7 +39,7 @@ You can read and write the file on booted target from linux:
    :substitutions:
 
    target:~$ cat /boot/bootenv.txt
-   overlays=conf-|dt-carrierboard|-peb-eval-01.dtbo#|dtbo-peb-av-10|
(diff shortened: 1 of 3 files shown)


Varakala02:
> Can you also check if it is required to adapt the remaining dtbo listings in the code-block (e.g. line 304 in pd24.1.0_mxp.rst)?
> 
It is not required here as these are actual devicetrees, the naming of the devicetree is not changed but only when assigning the devicetree to the overlays variable, a prefix is "conf-" is used.

> Similarly with the Devicetree substitutions e.g. line 67 in pd24.1.0_mxp.rst. I searched through the whole manual for "*.dtbo" to see at which places they have been used.
Actually for other devicetrees(dtbo) filenames, a prefix was already updated in the ../bsp/dt-overlays.rsti and included in the pd24.1.0_nxp.rst and head.rst.

jonas-rem (approved):

PR #255, opened 2024-11-18
source: yocto: Fix headline for own-layer-creation section

### LINKED ISSUE #253: Seems to be a cp error/leftover in heading
https://github.com/phytec/doc-bsp-yocto/blob/ae8601628278b0535b4e65d4d940c94b8ec0866b/source/yocto/kirkstone.rst?plain=1#L1508

s-hemer:
Did not fix the id in the translation file, nor the translated string.

Fixes #253

### DIFF
--- a/source/yocto/kirkstone.rst
+++ b/source/yocto/kirkstone.rst
@@ -1505,7 +1505,7 @@ to your *build/conf/local.conf*. Do not forget to rebuild the image
 
    host:~$ bitbake phytec-qt6demo-image
 
-Create your own Layer create layer
+Create your own layer
 ..................................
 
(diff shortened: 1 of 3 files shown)


jonas-rem (changes requested):
The translation is handled by the Phytec China team separately, so that we can ignore the change of the translation messages.

Can you remove the change of the translation message in this PR? It's better not to commit it, so that this change will show up when updating the translations.

s-hemer:
But that is just the 'msgid' which changed?! So it would break the translation. Agree, the 'msgstr' would then be the responsibility of Phytec CN but maybe the translation is already correct.

PR #263, opened 2024-12-09
fixes: fix issues found during translation review

jonas-rem:
This PR fixes some issues that have been found while reviewing the translation system.

- Move external ilnks to doc-bsp-yocto to sphinx references
- Reduce some unneded substitutions
- minor fixes that I saw on the way


annie-pixel (approved):
okay

PR #264, opened 2024-12-09
Update po files, containing translation messages

jonas-rem:
~Based on #263~

@annie-pixel I removed the comment lines in the po files, as those cause very large diffs at minimal changes of the content. Can you check if you observe any disadvantages when using the po files with poedit.


jonas-rem:
@annie-pixel update this PR after merging #263. Can you check it?

annie-pixel (approved):

Issue #254, opened 2024-11-16
Add hint to use/update/branch a new manifest at that point


s-hemer:
https://github.com/phytec/doc-bsp-yocto/blob/ae8601628278b0535b4e65d4d940c94b8ec0866b/source/yocto/kirkstone.rst?plain=1#L1445C1-L1458C34

It may make sense to highlight at this point, that this adding of additonal (own) layers should be represented/nailed-down in a manifest. 

Also this section has a lot of similarities to the following on adding an own layer.

PR #269, opened 2025-01-08
yocto: kirkstone: Add tip for creating own manifest file

s-hemer:
Suggestion for solving first part of #254


source/yocto/kirkstone.rst
```diff
@@ -1509,6 +1509,10 @@ to your *build/conf/local.conf*. Do not forget to rebuild the image
 
    host:~$ bitbake phytec-qt6demo-image
 
+.. tip::
+
+   After adding additional external or own layers, it makes sense to reflect those additions by deriving an own manifest file, cmp :ref:`kirkstone_phylinux-advanced-usage`
```
jonas-rem, line 1514: Can you reduce the number of characters per line to 80?

jonas-rem (changes requested):

source/yocto/kirkstone.rst
s-hemer: Will do, usually do not check this for github-only projects.

s-hemer:
Thanks for merging: yet, as it was a draft, the equivalent changes for scarthgap etc was missing/would be the same.

jonas-rem (approved):

s-hemer:
Looks like I was interrupted during the last round of fixes ... wanted to unset the draft directly afterwards.

s-hemer:
Just rebased to latest main

source/yocto/kirkstone.rst
```diff
@@ -1509,6 +1509,12 @@ to your *build/conf/local.conf*. Do not forget to rebuild the image
 
    host:~$ bitbake phytec-qt6demo-image
 
+.. tip::
+
+   After adding additional external or own layers, it makes sense to reflect
```
mschwan-phytec, line 1514: No need for "additional" here. That would be like saying "windy wind" or "wet rain".
```suggestion
   After adding external or own layers, it makes sense to reflect
```

s-hemer: Well, you are right. I thought about that as well. My idea why I left it was to highlight that this repeats after each additional layer but it is a duplication somehow.

jonas-rem (approved):

source/yocto/kirkstone.rst
svlstriker: what about using "further" instead of "additional"?

s-hemer: ok, if you insist. The longer I think about that "adding additional" the more I like and want to keep it.

svlstriker (approved):

PR #270, opened 2025-01-09
BSPIMX8M-3533 Wip/bhahn/add imx8mp libra

BHahn42618:
There will be a new phyCORE-i.MX 8M Plus FPSC SoM which will be utilizing the new Future Proof Solder Core (FPSC) standard. This module will be mounted on the Libra base board. As this will also use the imx8mp SoC, many features will be identical. 
Therefore copy the imx8mp files as a starting point. Then do some modifications that are needed for Libra and add the initial version of the manual to the build pipeline.

As many of the interfaces are not implemented / not working yet, I did not update them for the Libra. They need to be updated later, when they are implemented. Also the links to devicetrees still point to pollux and need to be updated later.  This PR creates only the structural base for the new Libra manual with some initial changes, that are already clear.


jonas-rem:
@BHahn42618 could you give some more context to the imx8mp-libra (in case this is information we can publicly share at the moment)?

ymoog:
> Add initial head manual for imx8mp-libra.

Please add some more details on what the scope of this PR is.

source/bsp/imx8/imx8mp-libra/imx8mp-libra.rst
```diff
@@ -0,0 +1,13 @@
+==========================
+i.MX 8M Plus Libra Manuals
```
tremmet, line 2: Add FPSC to the name here.

source/conf.py
```diff
@@ -207,6 +207,14 @@
         'manual',
         False,
     ),
+    (
+        'bsp/imx8/imx8mp-libra/head',
+        'imx8mp-libra-head.tex',
+        'i.MX 8M Plus Libra BSP Manual DRAFT',
```
tremmet, line 213: Add FPSC to the name here. too.

tremmet (changes requested):

ymoog (changes requested):
Directory and all other mentions should be named libra-imx8mp-fpsc or imx8mp-libra-fpsc to be consistent with either the "full" name or "software" name.
"Libra i.MX 8M Plus FPSC"
"imx8mp-libra-fpsc"

Other than that the changes that you do make look good.
I have not had a chance to look for changes that should have been made but weren't.

ymoog (approved):
Thanks!

PR #274, opened 2025-02-21
BSPIMX8M-3562 bsp: Add a known issue and workaround for SDK use

andrejpicej:
It was noticed that standalone u-boot/kernel build with SDK fails on some distros (e.g. Ubuntu20). The problem is that we are using the linker from SDK and linking against a mix of native and SDK libraries. The issues appear when some symbols were moved around between different .so files. In some versions the symbols are defined in dedicated libs and in newer library versions these symbols were defined in standard libc.

The simplest fix it to use native binutils, together with linker so lib dependencies are correctly followed.

For now this was only noticed on scarthgap NXP releases, where nativesdk-binutils is added to the SDK.


source/bsp/imx8/imx8mm/head.rst
```diff
@@ -263,6 +263,18 @@ To revert to the old style of booting, you may do
    device tree are missing in the boot partition.
 
 .. include:: /bsp/imx-common/development/standalone_build_preface.rsti
+
+.. warning::
+   Using SDK on some older host distros (Ubuntu 20), produces problems when building u-boot/kernel
```
tremmet, line 268: Can you specify the affected Yocto version in more detail here, too?
Same for the other entries.

tremmet (changes requested):

source/bsp/imx8/imx8mm/head.rst
andrejpicej: Yes sure, will add that this is a problem when using a SDK from Scarthgap NXP based BSPs.

tremmet (dismissed):

BHahn42618 (approved):

source/bsp/imx8/imx8mm/head.rst
```diff
@@ -263,6 +263,18 @@ To revert to the old style of booting, you may do
    device tree are missing in the boot partition.
 
 .. include:: /bsp/imx-common/development/standalone_build_preface.rsti
+
```
pfiser, line 266:
```suggestion
.. warning::  
   Using the SDK on older host distributions (e.g., Ubuntu 20.04 LTS) with Scarthgap NXP-based BSPs can cause issues  
   when building U-Boot or Linux kernel tools for host use. If you encounter an "undefined reference" error,  
   a workaround is to prepend the host's binutils to the PATH.  

   .. code-block:: console  
      host$ export PATH=/usr/bin:$PATH  

   Run this after sourcing the SDK *environment-setup* file.  
   
   Note, SDK issue has not been observed on newer distributions, such as Ubuntu 22.04, which appear to work 
   without requiring any modifications.

```

pfiser (commented):
Maybe rewrite a little bit?

source/bsp/imx8/imx8mm/head.rst
andrejpicej: Applied these changes :).

FauliSchlumpf (approved):

pfiser (approved):

PR #275, opened 2025-02-25
intl: update internationalization

jonas-rem:
Update the translation messages in the `.po` files to be in sync with the content.

Note:
This is a regular update of the translation messages. The content change slightly over time and smaller issues get fixed permanently. That means that we have to update the translation messages from time to time to be in sync.

This PR also fixes an issue that is currently causing the spellcheck action runner to fail (misspelled word "re-used").

@annie-pixel can you check with this PR if this makes it easy for you to update the translated strings in a follow-up PR? If you think that is the case, we can merge this and you can work on the translation update after that.


annie-pixel (approved):
okay

PR #276, opened 2025-02-25
Build: Add sphinx-autobuild as default for html build

jonas-rem:
I tested the [sphinx-autobuild](https://github.com/sphinx-doc/sphinx-autobuild) package with our documentation for some weeks now and found it to be very helpful while writing documentation. `sphinx-autobuild` automatically triggers an incremental html build, once a file in the filesystem has been changed. The html view is hosted at localhost and the browser view automatically refreshes on a finished build.

The `sphinx-autobuild` package is not part of the documentation, but is only used during the sphinx html build. The documentation can still be build in the regular way if preferred.

`py3 -e html` keeps the console open after a successful build and therefore `py3 -e intl` has been adapted and builds the html documentation for all languages without the `sphinx-autobuild` feature. It returns on success or failure and is used in the github actions from now on.

In addition this PR:

1. simplifies the tox build from `tox -e py3-html` to `tox -e html`
2. updates all python dependencies to their most recent version on `pypi`.

All those changes have been added to our documentation template already.


tox.ini
```diff
@@ -15,9 +15,6 @@ commands =
     sphinx-intl update -p build/locale -l zh_CN --line-width=10000
     sh -c './scripts/clean_po_files.sh'
     sphinx-build -E -W --keep-going -b html -D language=zh_CN source build/html/zh_CN -j auto
-
-[testenv:py3-html]
-commands =
```
ymoog, line 20: Please add this to the commit description. I got confused for a sec, because your description didn't mention this change.

ymoog (changes requested):

tox.ini
jonas-rem: These changes are related to the command that has to be used to build the documentation. The commit message describes this and gives an example.

```
build: tox: simplify build commands
Simplify tox build commands from e.g. "tox -e py3-html" to "tox -e
html".

Adapt github actions and documentation.
```


Could you suggest with an example what you would like to have added?

jonas-rem: Ah, thanks for explaining, got your point now! Yes, I will add an explanation it.

jonas-rem: The separation of changes within the commits should be much easier to understand now. It was indeed arranged in a very confusing way.

In addition I added the note to explain why it is not required anymore to call the "html" build command within the github workflows.

ymoog (approved):

tox.ini
ymoog: Thanks a lot, easy to understand now!

PR #278, opened 2025-03-06
linkcheck: fix permanently redirected links

jonas-rem:
Fix permanently redirected links in the documentation.

This PR also adds a script that I used to automatically fix redirects.


scripts/fix_redirects.py
```diff
@@ -0,0 +1,67 @@
...
+def get_redirects(linkcheck_output):
+    redirects = {}
+    redirect_pattern = re.compile(r'redirect\s+(http\S+)\s+-\s+permanently\s+to\s+(http\S+)')
+    for line in linkcheck_output:
+        match = redirect_pattern.search(line)
+        if match:
+            old_url = match.group(1)
+            new_url = match.group(2)
+            redirects[old_url] = new_url
+    return redirects
```
ymoog, line 29:
```suggestion
    redirect_pattern = re.compile(r'redirect\s+(http\S+)\s+-\s+permanently\s+to\s+(http\S+)')

    return {match.group(1): match.group(2) for line in linkcheck_output
            if (match := redirect_pattern.search(line))}
```

scripts/fix_redirects.py
```diff
@@ -0,0 +1,67 @@
...
+def update_links_in_file(file_path, redirects):
+    with open(file_path, 'r') as file:
+        content = file.read()
+
+    for old_url, new_url in redirects.items():
+        # Skip URLs ending with ".git"
+        if old_url.endswith('.git'):
+            continue
+        elif old_url in content:
+            content = content.replace(old_url, new_url)
```
ymoog, line 41:
```suggestion
        if not old_url.endswith('.git') and old_url in content:
            content = content.replace(old_url, new_url)
```

ymoog (approved):

scripts/fix_redirects.py
jonas-rem: Doesn't that make it harder to read? I don't really see the benefit of this change.

ymoog: Not to me. More concise and thus easier to understand what is happening.

ymoog (approved):

PR #279, opened 2025-03-13
Wip/bhahn/prep imx95 BSPIMX95-24

BHahn42618:
prepare imx95 alpha manual. Not all features are supported yet.


jremmet (commented):
I tested the video gstreamer examples and they work

tremmet (dismissed):

BHahn42618:
Added asjustments for ethernet and then rebased to main

source/bsp/imx9/imx95/components.rsti
```diff
@@ -0,0 +1,6 @@
+|sbc| Components
+................
+
+   **Libra FPSC Components (top)**
```
jonas-rem, line 4: Images are coming in a follow-up PR, right?

jonas-rem (changes requested):
> Not all features are supported yet.

I saw that you mentioned in the manual that the PEB-AV-20 is not supported yet. Are there any other features that are documented, but not yet supported? Or I guess missing features are just not added the manual yet?

The link to the release notes does not show the release notes yet.

Since this is an alpha release, documenting features in follow-up PRs should be no problem.

But a blocker is the currently broken linkcheck. If we merge this PR, the linkcheck will be broken for all other PRs as well. Do you plan to add the links to the website in the near future?

source/bsp/imx9/imx95/components.rsti
BHahn42618: I did not find any yet. I will have to go and ask the marketing if there are already Images of the libra baseboard.

jonas-rem:
> > I saw that you mentioned in the manual that the PEB-AV-20 is not supported yet. Are there any other features that are documented, but not yet supported? Or I guess missing features are just not added the manual yet?
> > The link to the release notes does not show the release notes yet.
> > Since this is an alpha release, documenting features in follow-up PRs should be no problem.
> > But a blocker is the currently broken linkcheck. If we merge this PR, the linkcheck will be broken for all other PRs as well. Do you plan to add the links to the website in the near future?
> 
> As the alpha release is not released yet and there is still ongoing development and new features beeing implemented, this manual does not reflect the state of a completed release, but the current development state (like the head manuals). We usually used the links of the last/latest release for the head manuals, but this is not possible for the imx95, as there is no previous release. I did change the links, so they do not point to imx8mp. So the links will probably continue not working until the alpha is released. You say this is a blocker, which I can fully understand. What is your opinion on how to solve it? I could think of creating a HACK/WIP branch or something like that and merging all the changes into that branch until the alpha is released or do you want to keep this PR open and merge PRs into this one until the alpha is released?

The option that you described would be fine, but then you do not have the alpha BSP documentation deployed during the development phase. One scenario is probably to have the documentation online so that interested parties can already work with the alpha BSP while it is being actively developed.

How about changing the links from `https://download.phytec.de/Software/Linux/BSP-Yocto-i.MX95/` to `https://download.phytec.de/Software/Linux/` so that the linkcheck does not fail. If people click on the link they will see the overview page and realize that the BSP is not yet released. This would be ok IMO. You could still have the full-length links as comment in the rst files and later replace the shortend links with the commented ones.

jonas-rem (approved):

PR #283, opened 2025-03-28
i.MX93 BSP documentation update for PD24.2.1

pfiser:
Update documentation for the i.MX93 PD24.2.1 release.

Changelog:
- update component version (Linux, U-Boot, Yocto, etc.)
- update line references in device-tree URL links
- rename SoM option extension overlays (unified with i.MX91)
- add phyboard-nash rev 1616.2
- improve WLAN/BT chapter for both boards (Segin uses PEB-WLBT-05, Nash uses PEB-WLBT-07)
- add mlan0 interface name note for PEB-WLBT-07 wireless adapter
- drop PEB-WLBT-05 specifics in general wireless-network.rsti chapter
- update CAN-FD chapter due to BSP changes
- update GPIO chapter due to BSP changes (libgpiod v2 was already part of previous release, so fix it now)
- update EEPROM chapter due to libgpiod v2 changes
- drop "BOOT/bootenv.txt" in favor of "/boot/bootenv.txt" in display.rsti chapter

PR #290, opened 2025-04-24
Add HEAD BSP manual for i.MX93

pfiser:
Add HEAD BSP manual for i.MX93 based on the latest PD24.2.1 release. It can be used to write documentation during BSP development cycle instead of at the end of the development cycle like we have done so far. This aligns the i.MX93 platform with other i.MX products from PHYTEC.

This was proposed by @jonas-rem in the PR #283.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -272,6 +272,14 @@ latex_documents = [
         False,
     ),
     (
+        'bsp/imx9/imx93/head',
+        'imx93-head.tex',
+        'i.MX 93 BSP Manual DRAFT',
+        'PHYTEC Messtechnik GmbH',
+        'manual',
+        False,
+    ),
+    (
         'bsp/imx9/imx93/pd24.1.0',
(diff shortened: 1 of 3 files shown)


jonas-rem (approved):

mschwan-phytec:
@cstoidner Could you do a quick review, so we can get this merged?

source/conf.py
```diff
@@ -271,6 +271,14 @@
         'manual',
         False,
     ),
+    (
+        'bsp/imx9/imx93/head',
+        'imx93-head.tex',
+        'i.MX 93 Mini BSP Manual DRAFT',
```
cstoidner, line 277: The word "Mini" above is a copy&paste fault, right?

cstoidner (changes requested):
Only the copy&paste error (see my comment) should be solved. The rest is fine for me.

cstoidner (approved):
Everything looks fine.

PR #298, opened 2025-05-07
BSPIMX95-72 imx95: add quickstart guide

BHahn42618:
Add quickstart guide for imx95 alpha BSP release. This guide only describes the first steps with the kit.


BHahn42618:
Will be integrated in https://github.com/phytec/doc-bsp-yocto/pull/313 to avoid merge conflicts.

PR #299, opened 2025-05-08
linkcheck: add some urls to ignore list

jonas-rem:
Github.com imposes rate limits after around 10 checks or so. It seems that this rate limit only counts for source file links, but not for general projects. For now we can ignore those files and instead manually checking those links from time-to-time with high wait times to not run into the rate-limit.

unix.stackexchange.com and jedec.org show other errors but the pages generally work. This is likely also recent measures to mitigate the impact of AI crawler.

Note:
This is a fix to make the linkcheck work reliably for other links. The situation seems to change rapidly with multiple websites recently and we may need to find other solutions for a linkcheck. First I would like to make the linkcheck work again (even if that means ignoring some domains). Then we can observe the behavior in the next weeks and adapt.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -374,6 +374,13 @@ latex_documents = [
 linkcheck_ignore = [
     "https://github.com/phytec/doc-bsp-yocto",
     "https://opencv.org/",
+    # Github rate limiting
+    "https://github.com/phytec/linux-phytec-imx",
+    "https://github.com/phytec/linux-phytec",
+    # Client Error 403: Forbidden
+    "https://unix.stackexchange.com/",
+    # Timeout with linkcheck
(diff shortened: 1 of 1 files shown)


jonas-rem:
@mschwan-phytec we can continue the discussion from https://github.com/phytec/doc-bsp-yocto/pull/291#issuecomment-2857351922 here.

I created a Personal Access Token, added that token to linkcheck and tested the authentication header with `mitmproxyq` locally. I saw the github token there, but github still did not counted those website requests against my quota.

I think the quota only works for the github API, not the regular webpage.

```
Authorization:    token github_pat_[..] 
```

Checking quota:
```
curl -H "Authorization: token github_pat_[..]" https://api.github.com/rate_limit
```

Another observation is that the quota seems to only apply to source files, not organisations or projects. My guess is that this is a recent change to defend against excessive AI crawlers.

PR #309, opened 2025-05-22
bsp: remove duplicate resizing ext4 chapter

jremmet:
The chapter is in
imx-common/emmc.rsti  -> using fdisk
peripherals/emmc.rsti  -> using parted

most manuals use both includes. Some only imx-common/emmc.rsti Removed the chapter from imx-common/emmc.rsti and add it to the manuals not using peripherals/emmc.rsti


source/bsp/imx8/imx8mm/pd22.1.1.rst
```diff
@@ -1781,6 +1781,8 @@ be read back as zeros.
 
 .. include:: /bsp/imx-common/emmc.rsti
 
+.. include:: /bsp/imx-common/emmc_resizing.rsti
```
mschwan-phytec, line 1784: I'd suggest using a more descriptive/fitting filename and use dashes to be consistent with other filenames.
```suggestion
.. include:: /bsp/imx-common/resize-rootfs.rsti
```

mschwan-phytec (changes requested):
Looks good overall. One small nit-pick:

mschwan-phytec (approved):
Thanks!

PR #311, opened 2025-05-23
bsp: imx9: imx93: head: Add doc-id information

pfiser:
Add information about document ID for i.MX93 BSP Manual. This aligns i.MX93 manual with other manuals for PHYTEC i.MX family of products.

### DIFF
--- a/source/bsp/imx9/imx93/head.rst
+++ b/source/bsp/imx9/imx93/head.rst
@@ -12,6 +12,7 @@
 .. _releasenotes: https://git.phytec.de/phy2octo/tree/releasenotes?h=imx93
 
 .. General Substitutions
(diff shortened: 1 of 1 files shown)


mschwan-phytec:
General question: Do we even use L-* document IDs for our Sphinx-based documentation? I thought this was previously used for the Confluence-based manuals, but should be irrelevant now.

pfiser:
Hi Martin, good question.
The missing doc-id came up when adding head.rst to i.MX93 manuals.
It was noted that it is missing for i.MX93, while all other in-tree i.MX platforms have it defined.
I think we should avoid the discrepancy. What do you think?

mschwan-phytec (approved):

PR #312, opened 2025-05-23
BSPIMX8M-3604 bsp: move e.MMC standard parts to common eMMC file

ymoog:
e.MMC boot partitions are defined as part of the e.MMC standard by JEDEC so they are not imx specific. Move them to the top generic file. Also, the JEDEC standard is called e.MMC, not eMMC. Adjust in the common files accordingly.


mschwan-phytec:
IMHO, the complete file `source/bsp/imx-common/emmc.rsti` could be moved to `source/bsp/peripherals/`, as the only i.MX specific things are the bootloader blob name "imx-boot", which could be replaced by a substitution. Any thoughts about this?

Name change to e.MMC as per JEDEC standard looks good to me.

ymoog:
> IMHO, the complete file `source/bsp/imx-common/emmc.rsti` could be moved to `source/bsp/peripherals/`, as the only i.MX specific things are the bootloader blob name "imx-boot", which could be replaced by a substitution. Any thoughts about this?

Sounds good. Makes sense since we would makes changes to all standard BSPs in case of boot partitions anyway.

source/bsp/peripherals/emmc.rsti
```diff
@@ -388,3 +388,111 @@ will be read back as zeros.
...
+~~~~~~~~~~~~~~~~~~~~~~
+
+On the host, run:
+
+.. code-block:: console
+
+   host:~$ scp <bootloader> root@192.168.3.11:/tmp/
+
+The partitions boot1 and boot2 are read-only by default. To write to them from
+user space, you have to disable force_ro in the sysfs.
```
mschwan-phytec, line 439:
```suggestion
user space, you have to disable ``force_ro`` in the sysfs.
```

source/bsp/peripherals/emmc.rsti
```diff
@@ -388,3 +388,111 @@ will be read back as zeros.
...
+The following table is for the offset of the |soc| SoC:
+
+===== ======================= =========== ==============
+SoC   Offset User Area        Offset Boot eMMC Device
+                              Partition
+===== ======================= =========== ==============
+|soc| |bootloader-offset| kiB 0 kiB       /dev/|emmcdev|
+===== ======================= =========== ==============
+
+After that set the boot partition from user space using the mmc tool:
```
mschwan-phytec, line 467:
```suggestion
After that set the boot partition from user space using the ``mmc`` tool:
```

source/bsp/peripherals/emmc.rsti
```diff
@@ -388,3 +388,111 @@ will be read back as zeros.
...
+
+To disable booting from the e.MMC boot partitions simply enter the following
+command:
+
+.. code-block:: console
+   :substitutions:
+
+   target:~$ mmc bootpart enable 0 0 /dev/|emmcdev|
+
+To choose back to the user area u-boot environment:
```
mschwan-phytec, line 493: This sentence makes little sense to me. Maybe the following is better?
```suggestion
To explicitly enable booting from the e.MMC user area, run:
```

mschwan-phytec (changes requested):

PR #317, opened 2025-06-10
imx93: add enabling jtag section

pefech:
JTAG isn't muxed by default on i.MX93.

This PR adds a section describing the need to enable it by devicetree overlay, because otherwise it does not work.

### DIFF
--- a/source/bsp/imx9/imx93/head.rst
+++ b/source/bsp/imx9/imx93/head.rst
@@ -390,6 +390,18 @@ Execute and power up the board:
 
 .. include:: ../../imx-common/development/format_sd-card.rsti
 
+Enabling JTAG Debug Interface
+-----------------------------
+
+The |soc| has a JTAG debug interface which can be used to debug software running
+on the SoC. The JTAG interface is routed through the pinmux and is not enabled
+by default. To enable the JTAG interface, please add this overlay to your
+```bootenv.txt``` file:
+
+.. code-block::
+
+   imx93-phyboard-nash-jtag.dtbo
+
 .. +---------------------------------------------------------------------------+
 .. DEVICE TREE
 .. +---------------------------------------------------------------------------+


cstoidner:
> 1. This step would be necessary for all releases, not only head, right?
> 

Actually not all. The overlay was introduced in PD24.2.0. So that part should be added to all release starting with PD24.2.0.

pefech:
I've added information about JTAG and MIPI_CSI sharing pins and I've added the JTAG section to all other PD24 manuals with the last push.

jonas-rem (approved):

jonas-rem:
> > ```
> > 2. You mentioned that the reason this overlay is not permanently activated was a pin conflict. Is that conflict described somewhere? If not, you could mention it here briefly.
> > ```
> 
> There is not really a pin conflict. It's just by nature that all JTag signals are on pins that can also mux other functionalities than the JTag signals. And as most Linux developers will not use JTAG, so we decide to not enable it by default.

Ok, thanks for the clarification. Do you think this short explanation would be helpful in the documentation? I feel it gives some context as to why this step is necessary and also serves as a hint to be aware of this and check for collisions in case other muxing options are used.

pefech:
I made a mistake and removed the section from PD24.1.x.
About the MIPI-CSI and JTAG pins: I put a note into the manual that clarifies that both interfaces are sharing pins, so only one interface can be used at time. I hope that wording is okay for you, @cstoidner.

jonas-rem (approved):

source/bsp/imx9/imx93/head.rst
```diff
@@ -390,6 +390,22 @@ Execute and power up the board:
...
+by default. To enable the JTAG interface, please add this overlay to your
+``bootenv.txt`` file:
+
+.. code-block::
+
+   imx93-phyboard-nash-jtag.dtbo
+
+.. note::
+   JTAG and the MIPI CSI interface cannot be used at the same time, since they
+   share the same pins.
```
cstoidner, line 407: Due to that conflict there is a jumper on the board to select which functionality should be served by those pins (see comment "JP15 is must be set if you want to use JTAG" in nash's the circuit diagram). Please add an corresponding comment to the BSP manual.

cstoidner (approved):
Look good now. Can be merged,

jonas-rem (approved):

PR #327, opened 2025-07-01
bsp: imx95-libra: remove USB-PD mention

ymoog:
on prototype hardware libra revision .0 and .1, usb-pd will pass the 24V supplied via phoenix connector back via the USB cable. To avoid users destroying their hardware, remove explanation chapter of USB-PD.

### DIFF
--- a/source/bsp/imx9/imx95/quickstart-alpha1.rst
+++ b/source/bsp/imx9/imx95/quickstart-alpha1.rst
@@ -189,13 +189,9 @@ kit. Connect the power supply to the Power-In connector (X8).
 
    .. figure:: /bsp/imx-common/images/power-connector.png
 
-It is possible to power the |sbc| with a USB-PD (USB power-delivery) supply
-instead of the power supply SV055. Connect a conform USB-PD supply to USB-C
-connector X2 right next to Power-In connector X8.
-
 .. warning::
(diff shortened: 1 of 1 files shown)


source/bsp/imx9/imx95/quickstart-alpha1.rst
```diff
@@ -189,13 +189,9 @@ kit. Connect the power supply to the Power-In connector (X8).
...
 
-It is possible to power the |sbc| with a USB-PD (USB power-delivery) supply
-instead of the power supply SV055. Connect a conform USB-PD supply to USB-C
-connector X2 right next to Power-In connector X8.
-
 .. warning::
-   Do not power the |sbc| via USB Power-In (X2) and Power-In (X8) at the same
-   time.
+   Do not power the |sbc| via USB Power-In, if populated, (X2) and Power-In (X8)
+   at the same time.
```
mschwan-phytec, line 194: I'd leave this sentence as is, to keep it clear and simple. It should be self-explanatory whether the USB power connector is populated or not.

(maybe just add "24V Power-In" to better distinguish from USB Power)

ymoog: Was instructed to phrase it this way.

PR #330, opened 2025-07-01
bsp: imx8: imx8mp-libra-fpsc: rework rs232

jremmet:


source/bsp/imx8/imx8mp-libra-fpsc/head.rst
```diff
@@ -326,54 +325,52 @@ Development
...
-configuration of the multi-protocol transceiver is done by jumpers |ref-jp3| and
-|ref-jp4| on the baseboard. For more information about the correct setup please
-refer to the |som|/|sbc| Hardware Manual section UARTs.
+The FPSC Standard supports 3 UART units. On the |sbc|, TTL level signals
+of UART3 (the standard console) and UART3 are routed to a FT4232H UART
+to USB converter expansion. This USB is brought out at USB-C connector X14.
+UART3 is connected to a multi-protocol transceiver for RS-232 and RS-485,
+available at pin header connector |ref-serial| at the RS-232 level,
+or at the RS-485 level. The muxing of the used transceivers is done by switch
+|ref-S5| on the baseboard.
```
ymoog, line 359: Should we mention that we can override S5 via devicetree? At least that is my observation so far, I haven't looked into the datasheets to see if this works reliably.

ymoog (approved):

source/bsp/imx8/imx8mp-libra-fpsc/head.rst
jremmet: Not sure how we want to use it
grafik
Looks like we can only force it to RS232 if we want to do this.
We can discuss this in the standup. Maybe add description if we provide overlays?

The kernel doesn't provide a general enable gpio support:
https://github.com/phytec/linux-phytec-imx/blob/v6.6.52-2.2.0-phy16/Documentation/devicetree/bindings/serial/rs485.yaml

PR #333, opened 2025-07-04
BSPIMX8M-3629 Wip/bhahn/bspimx8 m 3629 imx8mp libra eeprom

BHahn42618:
Add description for imx8mp-libra EEPROMs. Very similar to imx95.

### DIFF
--- a/source/bsp/imx-common/peripherals/eeprom.rsti
+++ b/source/bsp/imx-common/peripherals/eeprom.rsti
@@ -1,0 +1,63 @@
+EEPROM reading and writing in U-Boot
+....................................
+
+In U-Boot the ``i2c`` command can be used for EEPROM read and write operations.
+
+First set the correct i2c bus, where the eeprom is connected.
+`<bus-no>` is the i2c bus number of the EEPROM.
+
+.. code-block:: console
+
+   u-boot=> i2c dev <bus-no>
+
+To read and print the first 32 bytes from EEPROM execute.
+`<addr>` is the i2c address of the EEPROM.
+
+.. code-block:: console
+
+   u-boot=> i2c md <addr> 0x00 0x20
+
+To read the first 32 bytes from EEPROM into memory (loadaddr) execute:
+
+.. code-block:: console
+
+   u-boot=> i2c read <addr> 0x00 0x20 $loadaddr
+
+To write 0xff to the first 32 bytes execute:
+
+.. code-block:: console
+
+   u-boot=> i2c mw <addr> 0x00 0xff 0x20
+
+To write 32 bytes to EEPROM from memory (loadaddr) execute:
+
+.. code-block:: console
+
+   u-boot=> i2c write $loadaddr <addr> 0x00 0x20
+
+EEPROM reading and writing in Linux
+...................................
+
+Reading and writing can also be done via the sysfs in Linux. For this, find the
+correct path in sysfs first. It follows this logic:
+/sys/class/i2c-dev/i2c-<bus-no>/device/<bus-no>-<addr>/eeprom
+`<bus-no>` is the bus number of the EEPROM
+`<addr>` is the address of the EEPROM in 4 digits hex.
+
+To read and print the EEPROM content as a hex number, execute:
+
+.. code-block:: console
+
+   target:~$ hexdump -C <eeprom-sysfs-path>
+
+or:
+
+.. code-block:: console
+
+   target:~$ dd if=<eeprom-sysfs-path> | od -x
+
+To fill the EEPROM with zeros use:
+
+.. code-block:: console
+
+   target:~$ dd if=/dev/zero of=<eeprom-sysfs-path>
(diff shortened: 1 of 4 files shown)


source/bsp/imx-common/peripherals/eeprom.rsti
```diff
@@ -0,0 +1,63 @@
...
+....................................
+
+In U-Boot the i2c command can be used for EEPROM read and write operations.
+
+First set the correct i2c bus, where the eeprom is connected:
+
+.. code-block:: console
+   :substitutions:
+
+   u-boot=> i2c dev |i2c-bus-eeprom|
```
mschwan-phytec, line 11: IMHO, instead of using substitutions, adding a short section on how to find the correct I²C address of the EEPROM device is better. That would be much more versatile and applicable for all platforms in the future. The reader should have learned something after going through our manual, instead of just mindlessly copy-pasting commands.

We also don't want to maintain that much substitutions in the future. That is going to open up for errors/missing substitutions.

source/bsp/imx8/imx8mp-libra-fpsc/head.rst
```diff
@@ -452,22 +455,34 @@ General I²C2 bus configuration (e.g. |dt-carrierboard|.dts)
 EEPROM
 ------
 
-On the |som| there is an i2c EEPROM flash populated. It has two addresses. The
-main EEPROM space (bus: I2C-0 address: 0x51) can be accessed via the sysfs
-interface in Linux. The first 256 bytes of the main EEPROM and the ID-page
-(bus: I2C-0 address: 0x59) are used for board detection and must not be
-overwritten. Therefore the ID-page can not be accessed via the sysfs interface.
-Overwriting reserved spaces will result in boot issues.
+The system features three I2C EEPROM devices distributed across the SoM and carrier board:
```
mschwan-phytec, line 458: Please, wrap lines to 80 characters, as per our guidelines: https://github.com/phytec/doc-bsp-yocto/blob/main/CONTRIBUTING.rst#textwidth You may want to configure your editor accordingly...

source/bsp/imx8/imx8mp-libra-fpsc/head.rst
```diff
@@ -452,22 +455,34 @@ General I²C2 bus configuration (e.g. |dt-carrierboard|.dts)
...
+
+ * Board Detection EEPROM (write-protected)
+    * Bus: I2C-0
+    * Address: 0x51
+    * Purpose: Factory configuration for board identification
+
+ * User EEPROM
+    * Bus: I2C-0
+    * Address: 0x50
+    * Purpose: Available for user applications
```
mschwan-phytec, line 470: Wrong indentation and empty lines. Please align text to 3 spaces, as per our guidelines, and fix other instances, too. Again, I suggest you configure your editor accordingly.
```suggestion
*  Board Detection EEPROM (write-protected)
   *  Bus: I2C-0
   *  Address: 0x51
   *  Purpose: Factory configuration for board identification
*  User EEPROM
   *  Bus: I2C-0
   *  Address: 0x50
   *  Purpose: Available for user applications
```

source/bsp/imx-common/peripherals/eeprom.rsti
```diff
@@ -0,0 +1,63 @@
+EEPROM reading and writing in U-Boot
+....................................
+
+In U-Boot the i2c command can be used for EEPROM read and write operations.
```
mschwan-phytec, line 4:
```suggestion
In U-Boot the ``i2c`` command can be used for EEPROM read and write operations.
```

mschwan-phytec (changes requested):

source/bsp/imx-common/peripherals/eeprom.rsti
mschwan-phytec: > As far as I know there is no other way of finding out which EEPROM is a user EEPROM and which is a factory/reserved EEPROM, than to look into the documentation (maybe there is some comment in the devicetree source code of the EEPROM node, but no command that can easily be executed).

But how did you find out the correct address for this particular board in the first place? Is it because it's just the same as on the Pollux?

Do you mean the difference between an EEPROM on the module versus an EEPROM on the baseboard?

In any way, the module and baseboard schematic would definitely give away the correct address, right?

> These User EEPROMs do not necessarily need to have the same address for all boards. So I do not think we will get around a substitution here.

A simple table for all machines would also do the trick. No substitutions required. Then there would be also just one place to add different I²C addresses. And a table provides a nice overview when dealing with different machines.

BHahn42618: You mean doing it generic without telling a concrete bus / addr? Like this:
```suggestion
   u-boot=> i2c dev <i2c-bus-eeprom>

jremmet (dismissed):

source/bsp/imx-common/peripherals/eeprom.rsti
mschwan-phytec: I see, that's a good start. :+1:

PR #335, opened 2025-07-08
Wip/jremmet/lint

jremmet:
tox: add doc8 rst style checker
    
    What is checked:
        - invalid RST format - D000
        - lines should not be longer than 79 characters - D001
          - RST exception: line with no whitespace except in the beginning
          - RST exception: lines with http or https urls
          - RST exception: literal blocks
          - RST exception: rst target directives
        - no trailing whitespace - D002
        - no tabulation for indentation - D003
        - no carriage returns (use unix newlines) - D004
        - no newline at end of file - D005
    
    It used docutils to read the rst files and need a config to configure it
    Also set the project line length to 80 in tox.ini
    
    current result:
    
    Total files scanned = 51
    Total files ignored = 1
    Total accumulated errors = 406
    Detailed error counts:
        - doc8.checks.CheckCarriageReturn = 0
        - doc8.checks.CheckIndentationNoTab = 3
        - doc8.checks.CheckMaxLineLength = 403
        - doc8.checks.CheckNewlineEndOfFile = 0
        - doc8.checks.CheckTrailingWhitespace = 0
        - doc8.checks.CheckValidity = 0

tox: move single decency into tox
    
    This will hardly used outside


mschwan-phytec:
Very nice! :tada: This should make PRs easier to review.

I'll take a look...

jonas-rem:
I did not added anything like this in the past, because I wanted to avoid too many false positives. Users should not just blindly satisfy a linter.

But if a linter is configured accordingly it can be of great help and improve the documentation structure.

jremmet:
@jremmet 

> @jremmet Have you considered adding a GitHub Action for this, automatically checking PRs? Otherwise I think this check would be just ignored by users.

Yes, as I wrote ;) 
If we decide to use it, I'll try to fix the current long lines and add a github action.

jonas-rem:
> > I did not added anything like this in the past, because I wanted to avoid too many false positives. Users should not just blindly satisfy a linter.
> > But if a linter is configured accordingly it can be of great help and improve the documentation structure.
> 
> I agree, but too long lines and tab / spaces are hard to see in a review. Not sure it the 80 char are a hard requirement.

Right, this is hard to see and better done with a linter. No, it is not a hard requirement. At that time we (kind of) agreed that this improves readability while editing.

PR #348, opened 2025-07-29
BSPIMX8M-3674 add FPSC device tree concept documentation

ymoog:
For FPSC SoMs, rules set forth in the existing PHYTEC device tree concept are extended.
Document how FPSC is different to "normal" PHYTEC device tree concept.

### DIFF
--- a/source/bsp/fpsc-device-tree.rsti
+++ b/source/bsp/fpsc-device-tree.rsti
@@ -1,0 +1,108 @@
+PHYTEC |soc| FPSC Device Tree Differences
+.........................................
+
+Due to the standardization of signals and interfaces that PHYTEC FPSC SoMs must
+supply, there is potential for confusion of the names as defined by the SoC
+manufacturer, in Software by the device tree and in the FPSC specification.
+
+Example of i.MX 8M Plus SoC with a Libra i.MX 8M Plus FPSC board:
+
+========== =========== ==========
+SoC vendor device tree FPSC Spec.
+========== =========== ==========
+uSDHC1     usdhc1      SDIO
+uSDHC2     usdhc2      SD-Card
+uSDHC3     usdhc3
+========== =========== ==========
+
+To reduce confusion, interfaces in the device tree are annotated with the
+FPSC specification name.
+
+Example (taken from imx8mp-phycore-fpsc.dtsi)
+
+.. code-block:: devicetree
+
+   &usdhc2 { /* FPSC SDCARD */
+           pinctrl-0 = <&pinctrl_usdhc2>;
+           pinctrl-1 = <&pinctrl_usdhc2_100mhz>;
+           pinctrl-2 = <&pinctrl_usdhc2_200mhz>;
+           pinctrl-names = "default", "state_100mhz", "state_200mhz";
+           sd-uhs-sdr104;
+           vmmc-supply = <&reg_usdhc2_vmmc>;
+           vqmmc-supply = <&ldo5>;
+   };
+
+
+   pinctrl_usdhc2: usdhc2grp {
+           fsl,pins = <
+                   MX8MP_IOMUXC_SD2_CD_B__USDHC2_CD_B	0x40	/* SDCARD_CD */
+                   MX8MP_IOMUXC_SD2_WP__USDHC2_WP       0x40	/* SDCARD_WP */
+                   MX8MP_IOMUXC_SD2_CLK__USDHC2_CLK	0x190	/* SDCARD_CLK */
+                   MX8MP_IOMUXC_SD2_CMD__USDHC2_CMD	0x1d0	/* SDCARD_CMD */
+                   MX8MP_IOMUXC_SD2_DATA0__USDHC2_DATA0	0x1d0	/* SDCARD_DATA0 */
+                   MX8MP_IOMUXC_SD2_DATA1__USDHC2_DATA1	0x1d0	/* SDCARD_DATA1 */
+                   MX8MP_IOMUXC_SD2_DATA2__USDHC2_DATA2	0x1d0	/* SDCARD_DATA2 */
+                   MX8MP_IOMUXC_SD2_DATA3__USDHC2_DATA3	0x1d0	/* SDCARD_DATA3 */
+                   MX8MP_IOMUXC_GPIO1_IO04__USDHC2_VSELECT	0xc0
+           >;
+   };
+
+In the first node, the interface as defined per FPSC is annotated: ``FPSC SDCARD``
+In the second node, the signal names for this (SDCARD) interface are annotated:
+``SDCARD_WP``, etc. .
+
+FPSC Device Tree Structure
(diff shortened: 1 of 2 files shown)


source/bsp/fpsc-device-tree.rsti
```diff
@@ -0,0 +1,83 @@
...
+
+FPSC Device Tree Structure
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In addition to the usual PHYTEC device tree structure, all FPSC defined signals
+are also predefined in the SoM device tree as can be seen in the example above.
+There is nothing connected to the SDIO controller (on the SoM), yet the FPSC SoM
+must provide SDIO signals per the FPSC specification.
+The result is that on the baseboard, no care needs to be taken on whether the
+correct lines are connected; this is done already.
```
tremmet, line 53: I find  "correct lines are connected" a bit confusing as it has to be it is about configuration and not connecting. We can not connect anything. This needs to be done by hardware.

Maybe also add as example the SD-Card carrier board code. Examples are always better than text. This also helps to understand the phyCORE example below.

source/bsp/fpsc-device-tree.rsti
```diff
@@ -0,0 +1,83 @@
...
+
+Using the |som| for non-FPSC-conformant carrier boards
```
tremmet, line 55:
```suggestion
Using the |som| as phyCORE
```

source/bsp/fpsc-device-tree.rsti
```diff
@@ -0,0 +1,83 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Even though signals for pins are predefined within the FPSC world, the |som| may
+also be used in custom, non-conforming (with regards to FPSC) design.
```
tremmet, line 59: non FPSC  conform equals phyCORE.  Please mention it.

source/bsp/fpsc-device-tree.rsti
mschwan-phytec: I'm confused. Does "phyCORE" mean "violating FPSC standard"? Why do we have both in the same product name then: [phyCORE®-i.MX 95 FPSC](https://www.phytec.de/produkte/system-on-modules/phycore-imx-95-fpsc/)

mschwan-phytec: From a customer point of view "non-FPSC-conformant" exactly describes what you explained. So I think this is still better than saying "using as a phyCORE". The product name is _still_ "phyCORE", regardless of the standard being used.

The main question is rather: _Is the FPSC standard used or not?_ So it can only be one of "FPSC conformant" or "not FPSC conformant".

I would not introduce the term "phyCORE" (as an alternative to FPSC standard) here, as then "phyCORE" has too many meanings in different regards. ... which, again, is confusing.

tremmet: Well the phyCORE part was the marketing speech as how I understood it. So I would make use of it. But we can leave it like it's for now. This will need to be changed anyway again....and maybe I can get meanwhile some more clear information in how to name things properly from PM. And the non-FPSC-conformant Part is for sure a thing.

source/bsp/fpsc-device-tree.rsti
ymoog: I don't mind either phrasing, though it was my understanding that "multiplexing" as a concept means only one line is physically connected in hardware and the other lines are not.
So lines are connected based on the configuration in software.

tremmet: Yes, that is correct. The document is about software. This is why I expected more focus on the configuration than the connection.

ymoog: I understand that what is bothering you is that "connected" gives impression of something physical, is that correct?

I will change it to configuration.

source/bsp/fpsc-device-tree.rsti
ymoog: In this context, there is no phyCORE as the documentation talks about a carrierboard. If anything we would have to call it phyBOARD which does not make any sense either since customers probably don't name their carrier boards phyBOARD. For the instance below where the SoM is mentioned, I'll add the phyCORE keyword.

tremmet (dismissed):

source/bsp/fpsc-device-tree.rsti
```diff
@@ -0,0 +1,84 @@
...
+manufacturer, in Software by the device tree and in the FPSC specification.
+
+Example of i.MX 8M Plus SoC with a Libra i.MX 8M Plus FPSC board:
+
+========== =========== ==========
+SoC vendor device tree FPSC Spec.
+========== =========== ==========
+uSDHC      usdhc1      SDIO
+uSDHC      usdhc2      SD-Card
+uSDHC      usdhc3
```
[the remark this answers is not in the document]
ymoog: Yes, let's add the numbers.

source/bsp/fpsc-device-tree.rsti
ymoog: Fair point. I'll use SD card then.

source/bsp/fpsc-device-tree.rsti
```diff
@@ -0,0 +1,102 @@
...
+   &i2c4 {
+	pinctrl-0 = <&pinctrl_i2c4>;
+	pinctrl-names = "default";
+        status = "okay";
+   };
+
+   pinctrl_i2c4: i2c4grp {
+          fsl,pins = <
+                  MX8MP_IOMUXC_SD1_DATA3__I2C4_SDA 0x400001c2	/* I2C4_SDA */
+                  MX8MP_IOMUXC_SD1_DATA2__I2C4_SCL 0x400001c2	/* I2C4_SCL */
```
tremmet, line 96: To make it fitting to the now used examples above. i2c4 can be also muxed out of SD2:
```
MX8MP_IOMUXC_SD2_DATA0__I2C4_SDA
MX8MP_IOMUXC_SD2_DATA1__I2C4_SCL
```

tremmet (approved):

tremmet (dismissed):

BHahn42618 (approved):

jremmet (approved):

source/bsp/fpsc-device-tree.rsti
```diff
@@ -0,0 +1,102 @@
...
+========== =========== ==========
+uSDHC1     usdhc1      SDIO
+uSDHC2     usdhc2      SD-Card
+uSDHC3     usdhc3
+========== =========== ==========
+
+To reduce confusion, interfaces in the device tree are annotated with the
+FPSC specification name.
+
+Example (taken from imx8mp-phycore-fpsc.dtsi)::
```
mschwan-phytec, line 21: Have you considered using `code-block` for DTSI examples?

PR #367, opened 2025-09-24
bsp: am6x: Add phyCORE-AM62L FPSC platform

FPlohl:
This PR introduces documentation support for the phyCORE-AM62L FPSC platform.

### Changes

- Copied files that can be shared between TI and NXP platforms from `imx-common` to bsp/ directory to avoid duplication.
- Added `ti-common` directory for TI specific documentation.
- Added `linux-phytec-ti` link stub to `conf.py`.
- Created **HEAD BSP manual for AM62L** with relevant documentation included.

### Notes on Common Documentation

Some content of "common" documentation is **i.MX specific**. There are multiple solutions to this problem but we did not find an elegant way to solve this. Ideally we could conditionally include content based on platform (similar to `#ifdef` in C). The closest thing we found for RST is the `only` directive, but it works only at build time. That means we would have to build seperatly for NXP and TI, and then merge the documentation, which isn't practical.

One such problematic part is the NXP EULA which we currently excluded using `end-before`/`start-after` statements. 

We'd appreciate feedback on how to handle these cross-platorm documentation challenges more elegantly.

TO-DO:

- [ ] Set ${overlays} variable
- [ ] U-Boot External Environment
- [ ] Port all imx documentation to use new common include paths

### DIFF
--- a/source/bsp/am6x/am62l/head.rst
+++ b/source/bsp/am6x/am62l/head.rst
@@ -239,7 +239,7 @@ select the |som| default bootsource.
 .. include:: bootmode-switch.rsti
 
 .. include:: /bsp/ti-common/installing-os.rsti
-   :end-before: .. flash-spi-nor-flash-marker
+   :end-before: .. rauc-marker
 
 .. +---------------------------------------------------------------------------+
 .. DEVELOPMENT
--- a/source/bsp/ti-common/installing-os.rsti
+++ b/source/bsp/ti-common/installing-os.rsti
@@ -370,17 +370,41 @@ Flash SPI NOR Flash
 
 The |som|\ s are optionally equipped with SPI NOR Flash. To boot from SPI
 Flash, set |ref-bootswitch| to **SPI NOR**. The SPI Flash is usually quite small.
-The |kit| only has 32MB SPI NOR flash populated. Only the
-bootloader and the environment can be stored. The kernel, device tree, and file
-system are taken from e.MMC by default.
+The |kit| only has 64MiB SPI NOR flash populated. Only the bootloader and the
+environment can be stored. The kernel, device tree, and file system are taken
+from other boot sources.
 
-The SPI NOR flash partition table is defined in the U-Boot environment. It can
+The SPI NOR flash partition table is defined in the U-Boot device tree. It can
 be printed with:
 
 .. code-block::
 
-   u-boot=> printenv mtdparts
-   mtdparts=30bb0000.spi:3840k(u-boot),128k(env),128k(env:redund),-(none)
+   u-boot=> => mtd list
+   SF: Detected mt35xu512aba with page size 256 Bytes, erase size 4 KiB, total 64 MiB
+   List of MTD devices:
+   * nor0
+     - device: flash@0
+     - parent: spi@fc40000
+     - driver: jedec_spi_nor
+     - path: /bus@f0000/bus@fc00000/spi@fc40000/flash@0
+     - type: NOR flash
+     - block size: 0x1000 bytes
+     - min I/O: 0x1 bytes
+     - 0x000000000000-0x000004000000 : "nor0"
+             - 0x000000000000-0x000000080000 : "ospi.tiboot3"
+             - 0x000000080000-0x000000280000 : "ospi.tispl"
+             - 0x000000280000-0x000000680000 : "ospi.u-boot"
+             - 0x000000680000-0x0000006c0000 : "ospi.env"
+             - 0x0000006c0000-0x000000700000 : "ospi.env.backup"
+
+We have two possibilities for flashing:
+
+*  Flash SPI NOR with a combined binary, which contains all the necessary files
+   (tiboot3.bin, tispl.bin and u-boot.img) on the correct offsets. A whole
+   SPI NOR device in Linux is always marked as first MTD device (e.g.
+   /dev/mtd0) in U-Boot the device is called nor0.
+*  Flash SPI NOR with separate binaries to the corresponding partition. This can
+   be used when only certain bootloader part needs to be updated.
 
 Flash SPI NOR Flash from Network
 ................................
@@ -399,9 +423,8 @@ Flash SPI NOR from Network in kernel on Target
 *  Copy the image from the host to the target:
 
    .. code-block:: console
-      :substitutions:
 
-      host:~$ scp imx-boot-|yocto-machinename|-fspi.bin-flash_evk_flexspi root@192.168.3.11:/root
+      host:~$ scp ti-boot-container.img root@192.168.3.11:/root
 
 *  Find the number of blocks to erase of the U-boot partition:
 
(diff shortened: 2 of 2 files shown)


source/bsp/development/host_network_setup.rsti
```diff
@@ -150,15 +150,3 @@ When you boot/restart your host PC and don't have the network interface, as
...
-
-   DHCP server setup is only needed when using dynamic IP addresses. For our
-   vendor BSPs, static IP addresses are used by default.
-
-   .. code-block::
-
-      u-boot=> env print ip_dyn
-      ip_dyn=no
-
-   To use dynamic IP addresses for netboot, ``ip_dyn`` needs to be set to ``yes``.
```
ymoog: You are removing common code without replacement. Please give a reason why this is okay.

source/bsp/building-bsp.rsti
```diff
@@ -86,6 +86,7 @@ Starting the Build Process
       This needs to be done every time you open a new shell for starting builds.
 
 *  The current working directory of the shell should change to build/.
+.. nxp-eula-marker
```
ymoog, line 89: I think we should drop this and this should be a problem of i.MX manuals how they want to deal with this (i.e. informing readers of the FSL_EULA). The goal is to have the common code as generic as possible and this looks like an erroneous "forgotten" artifact.

source/bsp/ti-common/development/netboot.rsti
ymoog, line 1: What is the difference between the ti-common files and imx?
Is it possible to merge them into one generic rsti (for each ti-common files you introduce in this commit)?

To me it is important to know where the differences are. So that as a next step we can work on getting rid of the differences.

source/bsp/development/host_network_setup.rsti
andrejpicej: The idea behind this was that we now use dynamic IP for network boot with ampliphy-boot. I just recently noticed Benjamins u-boot patches after which we will probably support both. So I think this can be kept in. Also not sure if we now aggread that static IP or DHCP will be used in all releases?

source/bsp/am6x/am62l/bootmode-switch.rsti
dominiknh90, line 1: eMMC here is actually DEVBOOT which only got eFused to eMMC UDA. we need to update this for PD25 with the new hardware

source/bsp/am6x/peripherals/eeprom.rsti
```diff
@@ -0,0 +1,50 @@
...
+
+   target:~$ hexdump -c /sys/class/i2c-dev/i2c-0/device/0-0050/eeprom
+
+To read and print the first 1024 bytes of the EEPROM as a hex number,
+execute:
+
+.. code-block:: console
+
+   target:~$ dd if=/sys/class/i2c-dev/i2c-0/device/0-0050/eeprom bs=1 count=1024  | od -x
+
```
dominiknh90, line 23: this is configured as I2C1 by devicetree alias for FPSC

source/bsp/am6x/peripherals/tm.rsti
dominiknh90, line 1: we'll also have to make a split for imx common and ti common here

source/bsp/am6x/am62l/head.rst
```diff
@@ -0,0 +1,524 @@
...
+.. |link-boot-tools| replace:: https://download.phytec.de/Software/Linux/BSP-Yocto-AM62Lx/BSP-Yocto-Ampliphy-AM62Lx-ALPHA2/images/ampliphy-vendor/am62lxx-libra-fpsc-1/
+.. |link-bsp-images| replace:: https://download.phytec.de/Software/Linux/BSP-Yocto-AM62Lx/BSP-Yocto-Ampliphy-AM62Lx-ALPHA2/images/ampliphy-vendor/am62lxx-libra-fpsc-1/
+.. _releasenotes: https://git.phytec.de/phy2octo/tree/releasenotes?h=am62lx
+.. _`static-pdf-dl`: ../../../_static/am62l-fpsc-head.pdf
+
+.. AM62Lx specific
+
+.. General Substitutions
+.. |doc-id| replace:: Head
+.. |kit| replace:: **phyCORE-AM62L FPSC Kit**
```
dominiknh90, line 18: this will be phyFLEX-AM62L in the future

source/bsp/ti-common/development/netboot.rsti
dominiknh90: With the upcoming AM68 and AM69 in mind it makes sense for me to split it up since they can also load the bootloader over network.

source/bsp/ti-common/development/standalone_build_u-boot_binman.rsti
```diff
@@ -0,0 +1,120 @@
...
+
+.. code-block:: console
+   :substitutions:
+
+   host:~/|u-boot-repo-name|$ sudo dd if=flash.bin of=/dev/sd[x] bs=1024 seek=|bootloader-offset| conv=fsync
+
+.. hint::
+   The specific offset values are also declared in the Yocto variables
+   "BOOTLOADER_SEEK" and "BOOTLOADER_SEEK_EMMC"
+
```
dominiknh90, line 93: we have the ti-boot-container binary which is the 3 binaries glued together for this use case

source/bsp/ti-common/development/netboot.rsti
FPlohl: > What is the difference between the ti-common files and imx? Is it possible to merge them into one generic rsti (for each ti-common files you introduce in this commit)?
> 
> To me it is important to know where the differences are. So that as a next step we can work on getting rid of the differences.

Bulk of the content is pretty much the same so they can be merged into one generic rsti. A lot of the differences are not single line so we cannot use `replace` for platform specific documentation. 

We could use something like this:
- `generic.rsti` – contains the shared content with markers where platform differences occur.
- `imx.rsti` and `ti.rsti` – contain platform specific documentation and include `generic.rsti` with multiple `start-after/end-before` statements.
- In the board documentation we then include only the `imx.rsti` or the `ti.rsti`.

source/bsp/building-bsp.rsti
andrejpicej: Is this not needed anymore for newer NXP releases? Anyway, I see that this is also mentioned in yocto manuals.

leitzgen (approved):
Looks good.

source/bsp/building-bsp.rsti
ymoog: It is needed. What I am trying to convey is that this should be in source/bsp/<imx> not in /source/bsp for the sake of maintaining separation of common and imx specific parts.

source/bsp/ti-common/development/netboot.rsti
ymoog: I was hoping you could generate a diff easily. Anyway, I don't expect you to change this now.

source/bsp/peripherals/eeprom.rsti
```diff
@@ -1,17 +1,27 @@
+I2C EEPROM access
+.................
+
+.. warning::
+
+   The first 256 (0x100) bytes of the EEPROM area (bus: I2C-|eeprom-detect-bus|
```
dominiknh90: 256 bytes is not enough for api v3 with block extensions. we'll need a soc variable here (for am62l it's 1024 byte)

source/bsp/peripherals/eeprom.rsti
```diff
@@ -56,8 +66,38 @@ or:
...
+select the correct RAM configuration. This makes it possible to use the same
+bootloader image for different RAM sizes and choose the correct DTS overlays
+automatically.
+
+.. warning::
+
+   The first 256 (0x100) bytes of the EEPROM area (bus: I2C-|eeprom-detect-bus|
+   addr: |eeprom-detect-addr|) should not be erased or overwritten. As this will
+   influence the behavior of the bootloader. The board might not boot correctly
+   anymore.
```
dominiknh90: same as above

dominiknh90 (changes requested):
Hi Florijan,
I found a few more things to change.

In addition we'll have to rework the common eMMC page. IMX uses UDA with offset to boot and we'll have both UDA with FAT and HW boot partition (with offset 0) support in the PD25

source/bsp/am6x/am62l/head.rst
```diff
@@ -0,0 +1,517 @@
...
+USB Host Controller
+-------------------
+
+The USB controller of the |soc| SoC provides a low-cost connectivity solution
+for numerous consumer portable devices by providing a mechanism for data
+transfer between USB devices with a line/bus speed of up to 4 Gbit/s (SuperSpeed
+'SS'). The USB subsystem has two independent USB controller cores. Both cores
+are capable of acting as a USB peripheral device or a USB host. Each is
+connected to a USB 3.0 PHY.
+
```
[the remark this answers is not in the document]
FPlohl: Corrected this paragraph with accurate information

dominiknh90 (approved):
There are more open todos/fixes but I'm happy to address those in a new pull request as this one is already quite large.

source/bsp/development/master_manifest.rsti
```diff
@@ -0,0 +1,16 @@
+Accessing the Latest Upstream Support
+-------------------------------------
+
+We have a vanilla manifest that makes use of the Yocto master branches (not an
+TI release), Linux, and U-Boot. This can be used to test the latest upstream
```
BHahn42618: How about "vendor release)"? To make this general and not TI/NXP specific.

source/bsp/peripherals/network.rsti
```diff
@@ -0,0 +1,120 @@
+All interfaces offer a standard Linux network port that can be programmed using
+the BSD socket interface. The whole network configuration is handled by
+the systemd-networkd daemon. The relevant configuration files can be found on
+the target in ``/lib/systemd/network/`` as well as the BSP in
+``meta-ampliphy/recipes-core/systemd/systemd-conf``.
+
+IP addresses can be configured within \*.network files. The interfaces are
+configured to static IP as default. The default IP address and netmask for end0
```
BHahn42618: Is this end style something common for all BSPs, or something TI specific? Otherwise you could move this to ti-common folder.

source/bsp/ti-common/development/netboot.rsti
```diff
@@ -38,47 +33,42 @@ Place Images on Host for Netboot
...
+..............................
+
+Set the correct boot target, if not already set by u-boot board code and boot
+into network boot:
 
-.. Tip::
+.. code-block:: console
 
-   All supported devicetree overlays are in the |ref-dt| chapter.
+   u-boot=> setenv boot_targets dhcp
```
BHahn42618: Are you sure this is correct? I believe the boot_target for standardboot is ethernet not dhcp.

BHahn42618 (changes requested):

source/bsp/peripherals/eeprom.rsti
```diff
@@ -1,17 +1,27 @@
+I2C EEPROM access
+.................
+
+.. warning::
+
+   The first |eeprom-som-detect-area| (|eeprom-som-detect-area-hex|) bytes of
```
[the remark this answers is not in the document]
FPlohl: They should work. imx manuals don't include this file yet, they still use `source/bsp/imx9/peripherals/eeprom.rsti`, `source/bsp/imx8/peripherals/eeprom.rsti` or `/bsp/imx-common/peripherals/eeprom.rsti`

FPlohl:
I started working on AM68 documentation and I realised some changes needed to be made to common documentation. Notable changes: 
- moved `source/bsp/ti-common/gpio.rsti` to `source/bsp/peripherals/gpio.rsti` and moved board specific GPIO information to `am62l/head.rst`
- moved **Get the needed binaries** chapter from `source/bsp/ti-common/development/standalone_build_u-boot_binman.rsti` to `am62l/head.rst`
- moved `mtdinfo --all` code snippet from `source/bsp/am6x/peripherals/spi-nor-flash-no-boot.rsti` to `am62l/head.rst`

dominiknh90 (approved):

PR #384, opened 2025-12-08
BSPIMX95-140 TPM support

ymoog:
Introduce peripheral section to (Libra) i.MX 95 BSP manual that explains presence and basic usage with a TPM chip.

As a prerequisite, copy an adapted version of the i.MX 93 tpm section to the base bsp/peripherals.


Varakala02 (dismissed):

jremmet (commented):
Can we move bsp/imx9/imx91_imx93/peripherals/tpm.rst and reuse it in i.mx93 too?

I'm not sure if we want to left the copies all over. Something I also ask in #381

cstoidner:
> @cstoidner If you find a solution to make i.MX 9{1,3} include the generic tpm include, I'll add it in this PR (only if you even want to change the status quo).

I think if you use the |sbc| substitution ins1ead of the |kit|, it fits also for our imx91/93 platform. Although the |sbc| mentions the segin board also, but it's shorter form suggests more to the general imx93 family than to the specific carrier boards.

BHahn42618 (dismissed):

PR #391, opened 2026-01-08
Remove link checking

mschwan-phytec:
There are always URLs which are temporarily unavailable, old, do
not resolve correctly yet, etc. The more links are added in our
documentation and the older it gets, it is matter of time when we reach
a point where at least one link does not work at any given time. As a
result the GitHub workflow is broken for a long time and gets ignored.
The initial intention for checking every link was a good one, but at
this point does not provide us any benefit. In fact, the used scripts
are dead code by now. Because of this situation, I decided to remove the
checking altogether and make the GitHub workflow status for PRs
meaningful again.

Remove the unused script run_ci.sh. For executing checks, the
corresponding tox commands should be used directly.


ymoog (commented):
Would it not be useful to check for broken links? I.e. at least keep the script.
I agree, in practice it wasn't used. You could also decide to enforce the linkcheck and we would remove any links from "work-in-progress" manuals.

jremmet:
For manually checks `tox -e linkcheck` should be ok.
If I look into results there are also some worth to investigate: https://github.com/phytec/doc-bsp-yocto/actions/runs/20748921132/job/59572956597

@ymoog Removing all links from WIP - head versions make it hard to collect them when you finish a doc. I don't like delete partial information at all. We could remove all head docs and maintain a list of doc which should not be checked.

But in times of AI there are more and more counter measurements against scripts the access human readable links which results in random fails too.

mschwan-phytec:
> I agree that we should keep the script so that we can at least manually check the links before finishing up a new version.

> For manually checks `tox -e linkcheck` should be ok. If I look into results there are also some worth to investigate: https://github.com/phytec/doc-bsp-yocto/actions/runs/20748921132/job/59572956597

@ymoog @tremmet @jremmet But let's be honest: Would you actually use it, if the results are increasingly unreliable? Plus, being a manual step, I would guess this gets easily looked over.

> But in times of AI there are more and more counter measurements against scripts the access human readable links which results in random fails too.

This. It is increasingly difficult to automatically check links. There are just too many failure points to get reliable results that would justify such a checker, IMHO.

PR #414, opened 2026-02-11
Universal SVG DIP switches

mschwan-phytec:
Create universally usable DIP switch SVGs. These can be used for boot switches, RS485/RS232 mode switches and others.

For all BSP manuals, use SVGs for the boot switches instead of bitmap graphics.


ymoog (changes requested):
I think we can forfeit some of the realism for readability:
The positions of the individual switches were easier to read, I think partly because the contrast (white-black) was better and party because the vertical space between the high and low positions was more pronounced.

I suggest to either colour the switch knobs differently or to increase the distance. Maybe both.
Also, make the pictures themselves larger.

mschwan-phytec:
> I think we can forfeit some of the realism for readability: The positions of the individual switches were easier to read, I think partly because the contrast (white-black) was better and party because the vertical space between the high and low positions was more pronounced.

Thanks for the feedback! I already had the same thought. I'll revise the images and try a compromise between realism and readability.

> I suggest to either colour the switch knobs differently or to increase the distance. Maybe both. Also, make the pictures themselves larger.

Coloring the switch knobs seems a bit too much, IMHO. This might distract more from the actual look than necessary.

Increasing the scale of the images seems a good idea. 400% should do it.

mschwan-phytec:
> I'd use all of the horizontal space for the pictures (i.e. the table would be 100% text width) like it used to be. The size gets adjusted when viewing on a narrower screen so all images will always be viewable at the max possible width.

Scaling the images to a maximum looks odd, as the relation to the rest of the surrounding text is way off. This in turn, reduces text read flow. That's why I didn't scale them all the way up.

My goal is to keep it both readable, accessible and aesthetically pleasing to the eye.

Is the current 400% scale well readable for you?

> The changes on contrast and sizing of the switch images look good to me.

Okay, thanks!

ymoog:
> My goal is to keep it both readable, accessible and aesthetically pleasing to the eye.
> 
> Is the current 400% scale well readable for you?
> 

The size for the i.MX 95 looks good to me. For i.MX 93, I find it too small.
Any other opinions? @tremmet @s-hemer

mschwan-phytec:
> > My goal is to keep it both readable, accessible and aesthetically pleasing to the eye.
> > Is the current 400% scale well readable for you?
> 
> The size for the i.MX 95 looks good to me. For i.MX 93, I find it too small. Any other opinions? @tremmet @s-hemer

That is because the switch on the Libra board is also bigger in reality (because it is tri-state). The relative dimensions are equally scaled.

I am asking from an accessibility point of view: I'd say I personally have good sight, so I don't really have issues reading the switch knob states for neither of these. But of course, this may be different for other people. The knobs have similar sizes and dimensions in relation to the paragraph text. The indexes match that as well. I wonder, if the regular text is hard to read, too, if the knobs are difficult to distinguish? Or is it something different?

PR #443, opened 2026-04-09
bsp: generalize and update powerkey chapter

BHahn42618:
Append the powerkey chapter, with a check that the powerkey setting in systemd logind.conf is not overwritten by other configs. With the Yocto walnascar, the HandlePowerKey is deactivated by NXP in a seperate config which will overwrite the logind.conf. With that command the NXP config is shown.


tremmet:
Is this meant as a hint when something goes wrong? You do will fix this in the BSP, too?

ymoog (commented):
I don't understand the point of this addition.
Either our BSP does everything as described here, then we don't need the hint.
Or our BSP does not do this. Then we either change the doc or fix the BSP.

Where do you see value/justification in this addition?

BHahn42618:
In the previous BSP's NXP was editing the logind.conf directly via a patch. Now they deploy a seperate config file. The new way they do it seems to be the "upstream Yocto way", as this is the way how poky does it with `/usr/lib/systemd/logind.conf.d/00-systemd-conf.conf` (this was already present in the last BSP). The problem is, that the config file overwrites the logind.conf, which we describe should be changed. With this addition, the user should see that there are config files that overwrite the logind.conf. Since the new way seems to be the upstream way, I would not change it in Yocto and only add this to the BSP doc.
Also the config from NXP is needed, as not setting the button to ignore will lead to the Board shutting down when using the button to wakeup from suspend, which was the function we defined as the default in the past.

ymoog:
> In the previous BSP's NXP was editing the logind.conf directly via a patch. Now they deploy a seperate config file. The new way they do it seems to be the "upstream Yocto way", as this is the way how poky does it with `/usr/lib/systemd/logind.conf.d/00-systemd-conf.conf` (this was already present in the last BSP). The problem is, that the config file overwrites the logind.conf, which we describe should be changed. With this addition, the user should see that there are config files that overwrite the logind.conf. Since the new way seems to be the upstream way, I would not change it in Yocto and only add this to the BSP doc. 

I am positive that the BSP doc section you are referring to instructs to change the file at runtime, i.e. boot the system, edit the /etc/systemd/logind.conf file, save it. This has not changed.
I understand that the override happens within Yocto, is that correct?
Assuming the above, if we or anyone else downstream wants to change anything in the compile time contents of such a config file, a new file in our meta layer would have to be created anyway. This then takes precedence and we need not worry about anything NXP is doing.

> Also the config from NXP is needed, as not setting the button to ignore will lead to the Board shutting down when using the button to wakeup from suspend, which was the function we defined as the default in the past.

If I understood you correctly, we can take NXP version of the file, add our own flavour and save it in our meta layer.
I believe you mixed the runtime editing of the file with Yocto. If you want to add a hint, add it to the Yocto manual, not the BSP manual. But then again, using your own version of a config file is nothing new to Yocto, NXP just used to do it "incorrectly".

source/bsp/imx-common/peripherals/power-key.rsti
```diff
@@ -1,14 +1,14 @@
...
+Power Key
+---------
 
 The X_ONOFF pin connected to the ON/OFF button can be pressed long (for 5
 seconds) to trigger Power OFF without SW intervention or used to wake up the
-system out of suspend. With the *bbnsm_pwrkey* driver, the KEY_POWER event is
+system out of suspend. With the |powerkey-driver| driver, the KEY_POWER event is
 also reported to userspace when the button is pressed. On default, systemd is
 configured to ignore such events. The function of Power OFF without SW
 intervention are not configured. Triggering a power off with systemd when
```
tremmet:
```suggestion
intervention is not configured. Triggering a power off with systemd when
```
?

tremmet (dismissed):

BHahn42618:
rebase to main. imx91-93 pd26.1.0 manual was added

source/bsp/imx-common/peripherals/power-key.rsti
```diff
@@ -1,14 +1,14 @@
-bbnsm Power Key
----------------
+Power Key
+---------
 
 The X_ONOFF pin connected to the ON/OFF button can be pressed long (for 5
 seconds) to trigger Power OFF without SW intervention or used to wake up the
-system out of suspend. With the *bbnsm_pwrkey* driver, the KEY_POWER event is
+system out of suspend. With the |powerkey-driver| driver, the KEY_POWER event is
```
ymoog, line 6: Nitpick: Maybe better to use ``powerkey driver`` verbatim than introduce a substitution to reduce complexity.

```suggestion
system out of suspend. With the Power Key driver, the KEY_POWER event is
```

ymoog (approved):

tremmet (approved):

mschwan-phytec (approved):

PR #451, opened 2026-05-08
BSPIMX8M-4065: M Core updates

ymoog:
This PRs goal is to update the M Core section for i.MX8MP for PD26.1.0.

### DIFF
--- a/source/bsp/imx8/mcu.rsti
+++ b/source/bsp/imx8/mcu.rsti
@@ -89,19 +89,14 @@ page:
 Building the Firmware
 .....................
 
-To build the PHYTEC samples an environment has to be sourced
-
-.. code-block:: console
-
-   host:~/mcuxsdk$ source scripts/setenv
-
 The scripts to build the firmware are located in
-<sdk-directory>/phytec-mcux-boards/phyboard-pollux/<example_category>/<example>/armgcc.
+<sdk-directory>/examples-phytec/phyboard-pollux/<example_category>/<example>/armgcc.
 There are scripts for each memory location the firmware is supposed to run in,
 e.g.
 
 .. code-block:: console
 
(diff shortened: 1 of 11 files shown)


pefech:
Wouldn't it be more pretty, to rename the old file to something like _mcu.A0.rsti_ and put the content of the new mcu.rsti into a file called something like _mcu.head.rsti_?

Like this we would not have to paste the complete text in all older manuals.

ymoog:
> Wouldn't it be more pretty, to rename the old file to something like _mcu.A0.rsti_ and put the content of the new mcu.rsti into a file called something like _mcu.head.rsti_?
> 
> Like this we would not have to paste the complete text in all older manuals.

You'd have lots of A0, A1, A2, ... files. Also not that pretty; especially when you keep in mind that this rule would apply to all other shared content, even the one that changes frequently.

I like the paste rule we have now, imo it has more/better upsides than downsides.

source/bsp/imx8/mcu.rsti
BHahn42618: bootenv.txt is now named overlays.txt

source/bsp/imx8/mcu.rsti
```diff
@@ -173,9 +168,9 @@ adding |dtbo-rpmsg|:
 
    overlays=|dtbo-rpmsg|
```
BHahn42618: overlays variable is now named fit_overlay_conf

BHahn42618 (changes requested):

source/bsp/imx8/mcu.rsti
ymoog: Thanks, fixed it.

BHahn42618 (approved):

mschwan-phytec (approved):

PR #453, opened 2026-05-21
Wip/bhahn/imx8mp proof reeding pd26

BHahn42618:
Proofreading of imx8mp documentation for upcoming PD26.1.0. Mostly removing of unsupported features and some small fixes.

### DIFF
--- a/source/bsp/development/fitImages.rsti
+++ b/source/bsp/development/fitImages.rsti
@@ -27,8 +27,8 @@ Alternatively the content can also be printed on target in U-Boot:
    u-boot=> load mmc ${mmcdev}:1 ${loadaddr} fitImage
    u-boot=> iminfo ${loadaddr}
 
-Building fitImages standalone
-.............................
+Building FIT images standalone
+..............................
 
 For building a FIT, an image tree source file (``*.its``) and binaries to be packed are
 needed. The ``fitImage.its`` file and kernel and devicetree binaries can be
--- a/source/bsp/imx8/imx8mm/head.rst
+++ b/source/bsp/imx8/imx8mm/head.rst
@@ -181,24 +181,25 @@ First Start-up
 
 .. _imx8mm-head-images:
 
-* **u-boot.bin**: Binary compiled U-boot bootloader (U-Boot). Not the final
-  Bootloader image!
-* **oftree**: Default kernel device tree
-* **u-boot-spl.bin**: Secondary program loader (SPL)
-* **bl31-imx8mm.bin**: ARM Trusted Firmware binary
-* **lpddr4_pmu_train_2d_dmem_202006.bin,
-  lpddr4_pmu_train_2d_imem_202006.bin**: DDR PHY firmware images
-* **imx-boot**: Bootloader build by imx-mkimage which includes SPL, U-Boot, ARM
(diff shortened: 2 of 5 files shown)


mschwan-phytec (dismissed):

tremmet (dismissed):

source/bsp/imx8/imx8mp-fpsc/fpsc-device-tree.rsti
ymoog, line 1: My expectation is that the file will be made generic instead of moving it to imx8mp.
For 95 release, we have to move it back?!

Wouldn't it make more sense to keep it in the bsp directory?

BHahn42618: As half of the chapter consists of examples from the imx8mp devicetree and also the reset of the chapter was clearly designed with imx8mp-fpsc in mind and not imx95, I don't think, that it makes much sense to make it generic here. Instead I would create an extra chapter for imx95.
Also I don't know if the example with using SD interface pins as i2c works like this for imx95.

ymoog: After a closer look, I have to disagree. The chapter was clearly designed to be generic, you can see substitions already being present where appropriate.
Your changes destroy that concept. There are intentionally spelled out instead of using substitutions because they are toy examples.
Please leave this file as is in regards to this specific topic.

ymoog (changes requested):

source/bsp/imx8/imx8mp-fpsc/fpsc-device-tree.rsti
BHahn42618: I have to disagree with you here. This is not generic, as the chapter includes code example specific to one board: The old imx8mp Libra. As we did not change the code in the code examples, this happens to also fit for the new imx8mp Libra (except for the names), but the code examples do not fit the imx95.  I would not put this in the imx95 manual like this, as the imx8mp code examples would probably cause confusion with readers.

ymoog: I see your point. Note that this is labelled as an 'example'. That means that a specific implementation of a broader concept is given.

So I suppose your opinion is that we should not have a generic file with an example, rather each fpsc based manual should copy the same boilerplate and implement their own "correct" code showcase?
I prefer the example, so we need another opinion. @mschwan-phytec do you want to add your opinion?

mschwan-phytec: I agree with @ymoog here. It's fine to present examples for a specific platform and that example doesn't have to be exact nor match for other platforms. After all, we want to deliver the concept behind FPSC and not explain platform-specific implementations.

We do the same in our security and RAUC manuals, so this style of teaching/manual is done elsewhere, too.

@BHahn42618 Please keep the `fpsc-device-tree.rsti` at its current location and just include it as is. No need to use more substitutions.

BHahn42618: dropped the commit

tremmet (approved):

ymoog (approved):

mschwan-phytec (approved):

PR #463, opened 2026-06-23
bsp: {imx8,imx9}: Restore accidentally removed NXP EULA note

cstoidner:
Commit 8aea814c11aefb0a97753aa4481d6e061d3cc199 tried to move NXP EULA note from the general documentation file to NXP specific documentations. But the commit just removed the EULA note.

Restore the NXP EULA note for the NXP based vendor releases.

### DIFF
--- a/source/bsp/imx8/imx8mm/head.rst
+++ b/source/bsp/imx8/imx8mm/head.rst
@@ -178,6 +178,19 @@ First Start-up
 .. +---------------------------------------------------------------------------+
 
 .. include:: /bsp/building-bsp.rsti
+   :end-before: .. nxp-eula-marker
+*  Open the main configuration file and accept the GPU and VPU binary license
+   agreements. Do this by uncommenting the corresponding line, as below.
+
+   .. code-block:: console
+
+      host:~/yocto/build$ vim conf/local.conf
+      # Uncomment to accept NXP EULA
+      # EULA can be found under ../sources/meta-freescale/EULA
+      ACCEPT_FSL_EULA = "1"
+
+.. include:: /bsp/building-bsp.rsti
+   :start-after: .. nxp-eula-marker
 
 .. _imx8mm-head-images:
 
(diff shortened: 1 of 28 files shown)


mschwan-phytec:
Thanks for fixing this!

Maybe it's worth putting this note into its own include file and just use that instead of copying the text many times.

ymoog (changes requested):
We need the note in all i.MX manuals I believe. Please add the note to the imx8mp mainline-head manual
At least the EULA is present in imx-firmware.

krstruzyna:
> We need the note in all i.MX manuals I believe. Please add the note to the imx8mp mainline-head manual At least the EULA is present in imx-firmware.

Isn't the EULA only needed on vendor-specific releases? Mainline shouldn't need it if I understand it correctly.

cstoidner:
> We need the note in all i.MX manuals I believe. Please add the note to the imx8mp mainline-head manual
> At least the EULA is present in imx-firmware.

@ymoog , which firmware do you mean? The ddr firmware used in u-boot? Or anything else?

ymoog:
> > We need the note in all i.MX manuals I believe. Please add the note to the imx8mp mainline-head manual
> > At least the EULA is present in imx-firmware.
> 
> @ymoog , which firmware do you mean? The ddr firmware used in u-boot? Or anything else?

firmware-ele-imx, firmware-imx for example.

cstoidner:
> > > We need the note in all i.MX manuals I believe. Please add the note to the imx8mp mainline-head manual
> > > At least the EULA is present in imx-firmware.
> > 
> > 
> > @ymoog , which firmware do you mean? The ddr firmware used in u-boot? Or anything else?
> 
> firmware-ele-imx, firmware-imx for example.

Understood. I will take care.

krstruzyna (approved):

ymoog (approved):

PR #467, opened 2026-07-07
imx95-fpsc: update thermal management chapter

BHahn42618:
The imx95 will likely never feature thermal management/protection support in U-Boot, so remove that chapter. Add a chapter for imx-sm instead. This feature/behaviour has not been tested for the alpha releases, so do not include it in alpha manuals.

### DIFF
--- a/source/bsp/imx9/imx95-fpsc/alpha2.rst
+++ b/source/bsp/imx9/imx95-fpsc/alpha2.rst
@@ -531,7 +531,81 @@ Device tree description of LVDS-1 can be found here:
 
 .. include:: /bsp/imx8/peripherals/pm.rsti
 
-.. include:: peripherals/tm.rsti
+Thermal Management
+------------------
+
+U-Boot
+......
+
+There is no Thermal Management support in the first ALPHA release for the
+i.MX95 in U-Boot.
+
+Kernel
+......
+
+The Linux kernel has integrated thermal management that is capable of monitoring
+SoC temperatures, reducing the CPU frequency, driving fans, advising other
+drivers to reduce the power consumption of devices, and – worst-case – shutting
+down the system gracefully
+(https://www.kernel.org/doc/Documentation/thermal/sysfs-api.txt).
+
+This section describes how the thermal management kernel API is used for the
+|soc| SoC platform.
+
+There are nine temperature sensors on the SoM that are readable from Linux.
+The |socfamily| has two internal temperature sensors for the SoC.
+Three internal sensors for the PMICs and four I²C temperature sensors located
+close to DRAM, eMMC, ethernet PHY and PMIC.
+
+*  The current temperatures of the system can be read in milli celsius over
+
+   .. code-block:: console
+
+      target:~$ cat /sys/class/hwmon/hwmon*/temp*_input
+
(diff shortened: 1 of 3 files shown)


source/bsp/imx9/imx95-fpsc/peripherals/tm.rsti
```diff
@@ -73,3 +67,23 @@ default policy used in the BSP is step_wise.
 
    If the value of the SoC temperature in the sysfs file temp reaches
    *trip_point_1* (critical), the board immediately shuts down to avoid any heat damage.
+
+.. imx95-tm-imx-sm-marker
+
+Systemmanager (imx-sm)
```
ymoog, line 71: Maybe it would be better to create an include dir for system manager?
Behaviour is different to "normal" thermal mgmt.

BHahn42618: This is a imx95-fpsc thermal management chapter not a general imx chapter!

ymoog: Correct. My suggestion is to move it out of the imx95-fpsc directory.

BHahn42618: How do you imagine this? This chapter is only for imx95. Thermal management is different for imx95 than other SoCs, as imx-sm plays a major part in it. The other SoCs we use do not have imx-sm (like imx8), so I do not see this beeing a chapter that we can or should generalize and move out of imx95 directory.

ymoog: i.MX 952 does have SM and we can reuse it. Maybe the generation after i.MX 9 will also use SM. That is my motivation for the suggestion.
As to how I image this:
We create a new directory called e.g. system-manager and put a thermal management include file in there. All SoCs with SM can then include files in the system manager directory and it is clearly marked as separate from Linux kernel.

tremmet: I think a system manager folder with different chapter(-parts) will it make quite messy. 

Please dissolve the thermal chapter for the alpha releases and remove the marker. When we have more platforms that use SM we can make it more generic if it is reusable.

source/bsp/imx9/imx95-fpsc/alpha2.rst
```diff
@@ -531,7 +531,81 @@ Device tree description of LVDS-1 can be found here:
...
 .. include:: /bsp/imx8/peripherals/pm.rsti
 
-.. include:: peripherals/tm.rsti
+Thermal Management
+------------------
+
+U-Boot
+......
+
+There is no Thermal Management support in the first ALPHA release for the
```
tremmet: second ALPHA :)

tremmet (changes requested):

source/bsp/imx9/imx95-fpsc/peripherals/tm.rsti
ymoog: I am okay with just having an extra file in the current directory, then. E.g. sm-tm.rsti

